PR #1, opened 2023-06-15
Readme update

tmichela:
@JamesWrigley just quickly updating the README and a few urls

I realize how messed up is the whole naming 🙈


JamesWrigley (approved):
Ah nice :laughing: LGTM!

PR #2, opened 2023-06-16
Migrate CI to GitHub workflow

tmichela:


.github/dependabot/constraints.txt
```diff
@@ -0,0 +1,20 @@
+euxfel_bunch_pattern
```
JamesWrigley, line 1: I think these should also have version numbers, e.g. from `pip freeze`.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,47 @@
...
+        packages: ${{ env.SYSTEM_PACKAGES }}
+
+    - uses: actions/cache@v3
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('**/setup.py') }}-${{ hashFiles('.github/dependabot/constraints.txt') }}
+
+    - name: Install Python dependencies
+      run: |
+        python3 -m pip install --upgrade pip
```
JamesWrigley, line 40: Why is this is necessary? (not saying it should be removed, just curious)

.github/workflows/tests.yml
```diff
@@ -0,0 +1,47 @@
...
+        python3 -m pip install .[test] --constraint .github/dependabot/constraints.txt
+
+    - name: Test with pytest
+      run: |
+        python3 -m pytest -v
+      env:
+        MPLBACKEND: agg
```
JamesWrigley, line 47: Is this necessary? I don't think we needed it in the Gitlab CI.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    - name: Cache APT Packages
+      uses: awalsh128/cache-apt-pkgs-action@v1.3.0
+      with:
+        packages: ${{ env.SYSTEM_PACKAGES }}
+
+    - uses: actions/cache@v3
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('**/setup.py') }}-${{ hashFiles('.github/dependabot/constraints.txt') }}
```
JamesWrigley, line 36: The first `hashFIles()` should be pointing to `pyproject.toml`, we don't have a `setup.py`.

JamesWrigley (changes requested):
How annoying would it be to rollback the big rename? Asking because it'll probably be a bit painful rebasing the existing branches :sweat_smile: (yes, I'm aware this is entirely my fault)

JamesWrigley:
Oh, and we can also get rid of `.gitlab-ci.yml` now.

.github/workflows/tests.yml
tmichela: left over from copying the yaml file (was necessary in extra-data, but can't recall why)

tmichela:
> How annoying would it be to rollback the big rename?

What do you suggest we do? Because I don't see the branches disapear anytime soon

JamesWrigley (approved):
I would suggest leaving the renaming till later if it's not too work much to revert. But I leave it to you, I don't feel very strongly about it. Other than that, LGTM!

tmichela:
Clarified offline with James, leaving the renaming 📛 
Thanks for the review!

PR #9, opened 2023-06-21
Dialog to select proposal/directory when starting GUI 

takluyver:
image

I haven't tried to do any refactoring to take full advantage of this yet.

### DIFF
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -199,9 +199,6 @@ da-dev@xfel.eu"""
 
         self.autoconfigure(context_dir, proposal=prop_no)
 
-    def gpfs_accessible(self):
-        return os.path.isdir("/gpfs/exfel/exp")
-
     def save_settings(self):
         self._settings_db_path.parent.mkdir(parents=True, exist_ok=True)
 


damnit/gui/open_dialog.py
```diff
@@ -0,0 +1,74 @@
...
+        self.ui.buttonBox.button(QDialogButtonBox.StandardButton.Ok).setEnabled(valid)
+
+    def browse_for_folder(self):
+        path = QFileDialog.getExistingDirectory()
+        if path:
+            self.ui.folder_edit.setText(path)
+
+    def get_chosen_dir(self):
+        if self.ui.proposal_rb.isChecked():
+            return Path(self.proposal_dir, "usr/Shared/amore")
```
JamesWrigley, line 64: Nitpicking:
```suggestion
            return Path(self.proposal_dir) / "usr/Shared/amore"
```

JamesWrigley (changes requested):
Things we discussed:
- Adding tests
- Open dialog from menu entry
- Creating a backend from the dialog if one doesn't exist

takluyver:
Thanks @JamesWrigley for the review. I've made all the changes we discussed now; do you want to take another look overt the latest changes? The main change, other than the tests, is to the code for creating a DAMNIT DB and starting the backend if they don't already exist.

JamesWrigley:
Hmmm, now I'm getting crashes when testing this on Maxwell and my local machine :thinking: Both with selecting by proposal and directory:
```
(amore) ➜  mid-2832 git:(open-db-dialog) amore-proto gui --no-kafka    
QThread: Destroyed while thread is still running
[1]    3384071 IOT instruction (core dumped)  amore-proto gui --no-kafka
```

takluyver:
Gah, threads. I saw those errors while running the tests and fixed them there, but then didn't try again with the actual application. I'll take a look.

takluyver:
OK, I think I've finally got the application, the local tests and the CI all working :sweat_smile: :crossed_fingers: .

JamesWrigley (approved):
I didn't have quite enough time to test it myself, but I trust you and CI :grin: LGTM!

(P.S. I might also suggest squashing some of the commits before merging, but I leave that up to you)

PR #20, opened 2023-07-05
Fix the instructions for connecting to XFELs broker offline

JamesWrigley:
I'm not sure how this worked originally, but when testing it I found that the KafkaConsumer will connect to the bootstrap server (localhost) correctly with the AMORE_BROKER variable set, but it will then try to connect to exflwebstor01.desy.de and hang. I'm guessing that when it connects to the bootstrap server it gets a list of all brokers and then tries to connect to them; and since it sees localhost and exflwebstor01.desy.de as different brokers it'll try to connect to exflwebstor01.desy.de too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ at the command line.
 ## Kafka
 The GUI is updated by Kafka messages sent by the backend. Currently we use
 XFEL's internal Kafka broker at `exflwebstor01.desy.de:9102`, but this is only
-accessible inside the control network.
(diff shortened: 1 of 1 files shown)


takluyver:
> this is only accessible inside the control network.

I think we mean inside the DESY network? The control network is more restricted - I don't think it includes Maxwell.

Other than that, LGTM.

JamesWrigley:
Oops, yes you're quite right. Fixed in 30095dd.

PR #21, opened 2023-07-05
Slurm improvements

JamesWrigley:
See the commit messages for details, but TL;DR:
- Makes the slurm timeout configurable (and default to 2h)
- Save the slurm logs in a dedicated directory and give the logs and jobs a readable name


takluyver:
Assuming you've tested this, LGTM. :slightly_smiling_face:

PR #22, opened 2023-07-05
Set the aspect ratio of images automatically based on their shape

JamesWrigley:
Anything with an aspect ratio >4 gets a dynamic aspect ratio by default.

Example:
damnit-aspect-ratio

### DIFF
--- a/damnit/gui/plot.py
+++ b/damnit/gui/plot.py
@@ -3,6 +3,7 @@ import numpy as np
 import pandas as pd
 from pandas.api.types import is_numeric_dtype
 
+from PyQt5.QtCore import Qt
 from PyQt5 import QtCore, QtWidgets
 from PyQt5.QtWidgets import QMessageBox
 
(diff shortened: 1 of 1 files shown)


damnit/gui/plot.py
```diff
@@ -217,6 +232,11 @@ def update_canvas(self, xs=None, ys=None, image=None, legend=None, series_names=
             vmin = np.nanquantile(image, 0.01, interpolation='nearest')
             vmax = np.nanquantile(image, 0.99, interpolation='nearest')
             self._image.set_clim(vmin, vmax)
+
+            # Take a guess at a good aspect ratio
+            aspect_ratio = max(image.shape[:2]) / min(image.shape[:2])
+            if aspect_ratio > 4:
+                self._dynamic_aspect_checkbox.setCheckState(Qt.Checked)
```
takluyver, line 239: I think calling it inside `.update_canvas()` means it could automatically check itself as new data arrives, though I haven't gone through to work out if an image plot can ever be automatically updated.

JamesWrigley: Fair enough, fixed in 92efa3e. Maybe we will add support for updating images in the future (or at least warning a user if an image has been updated).

takluyver:
Thanks, LGTM

PR #23, opened 2023-07-05
Allow setting default arguments for dependencies

JamesWrigley:

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -7,6 +7,7 @@ possibly running in a different Python interpreter. It will run a context file
 
 import argparse
 import functools
+import inspect
 import logging
 import os
 import time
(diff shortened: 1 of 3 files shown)


damnit/ctxsupport/ctxrunner.py
```diff
@@ -211,16 +212,28 @@ def __init__(self, data, ctx):
...
         for name in ctx_file.ordered_vars():
             var = ctx_file.vars[name]
 
             try:
                 # Add all variable dependencies
-                kwargs = { arg_name: res.get(dep_name)
+                kwargs = { arg_name: get_dep_or_default(var, arg_name, dep_name)
                            for arg_name, dep_name in var.arg_dependencies().items() }
 
                 # If any are None, skip this variable since we're missing a dependency
```
takluyver, line 235:
```suggestion
                # Check for missing dependencies with no default value in the function
```

Since we're no longer looking for None. :slightly_smiling_face:

takluyver:
Other than that comment fix, LGTM

damnit/ctxsupport/ctxrunner.py
JamesWrigley: Hah, indeed :) Fixed in 6085018.

PR #24, opened 2023-07-05
Add support for exporting the table to a CSV/Excel file

JamesWrigley:


JamesWrigley:
(marking this as a draft because I'm not happy with the tests yet)

damnit/util.py
```diff
@@ -14,3 +23,11 @@ def wait_until(condition, timeout=1):
 
     if slept_for >= timeout:
         raise TimeoutError("Condition timed out")
+
+def timestamp2str(timestamp):
+    if timestamp is None or pd.isna(timestamp):
+        return None
+
+    dt_naive = datetime.fromtimestamp(timestamp)
+    dt_local = dt_naive.replace(tzinfo=timezone.utc).astimezone()
```
takluyver, line 32:
```suggestion
    dt_utc = datetime.fromtimestamp(timestamp, timezone.utc)
    dt_local = dt_utc.astimezone()
```

`fromtimestamp()` without a timezone actually gives you a naive datetime representing *local* time, which `.replace()` then labels as UTC, so you double the offset. Compare:

```
In [12]: datetime.fromtimestamp(ts).replace(tzinfo=timezone.utc).astimezone().strftime("%H:%M:%S %d/%m/%Y")
Out[12]: '19:05:24 07/07/2023'

In [13]: datetime.fromtimestamp(ts, timezone.utc).astimezone().strftime("%H:%M:%S %d/%m/%Y")
Out[13]: '18:05:24 07/07/2023'
```

My local clock was 18:05 when I did `ts = time.time()`, and that's 17:05 in UTC. The first version puts me 2 hours ahead instead of 1.

You could also just leave it with the naive local datetime and format that:

```python
dt_local = datetime.fromtimestamp(timestamp)
return dt_local.strftime("%H:%M:%S %d/%m/%Y")
```

But I like to make datetimes timezone-aware wherever possible. :slightly_smiling_face: 
```

File: damnit/gui/main_window.py [hunk 2]
```diff
@@ -499,6 +509,52 @@ def _create_menu_bar(self) -> None:
...
+        columns = ["Status"] + list(self.table_view.get_column_states().keys())
+        cleaned_df = self.data[columns].copy()
+        # Format timestamps nicely
+        cleaned_df["Timestamp"] = cleaned_df["Timestamp"].map(lambda x: timestamp2str(x))
+        # Format images nicely
+        cleaned_df = cleaned_df.applymap(image2str)
+
+        if extension == ".xlsx":
+            proposal = self.db.metameta["proposal"]
+            cleaned_df.to_excel(export_path, sheet_name=f"p{proposal} DAMNIT run table")
```
takluyver, line 551: Getting on my FOSS horse :horse: : if we're doing spreadsheet export, can we support OpenDocument as well as Excel? Weirdly this seems to go through the `.to_excel()` method with an `engine` parameter to say that it's actually ODS:

https://pandas.pydata.org/docs/user_guide/io.html#opendocument-spreadsheets

This would require the [odfpy](https://pypi.org/project/odfpy/) package as well.

I haven't tested this, but it looks like the additional code should be like this:

```suggestion
        if extension == ".xlsx":
            proposal = self.db.metameta["proposal"]
            cleaned_df.to_excel(export_path, sheet_name=f"p{proposal} DAMNIT run table")
        elif extension == ".ods":
            proposal = self.db.metameta["proposal"]
            cleaned_df.to_excel(export_path, engine="odf", sheet_name=f"p{proposal} DAMNIT run table")
```

damnit/util.py
JamesWrigley: You're quite right, but now I'm massively confused because the current code is supposedly giving the right timestamps :upside_down_face: Taking p3303 as an example, with the current code run 1 is supposed to have started at 1:14PM:
image

Which according to [mymdc](https://in.xfel.eu/metadata/proposals/819/runs/176414) is correct:
image

If I try with your code I get a start time of 12:14PM instead, an hour earlier. So there's definitely an extra hour being added, but does that mean that the time in mymdc is incorrect? I can't see any other reason for the discrepancy, your explanation makes sense to me.

damnit/gui/main_window.py
JamesWrigley: I tried this out, but for some reason the boolean `Status` column isn't handled properly :frowning_face: 
image

The odfpy library also seems kinda abandoned, so I'm leaning towards not supporting ODS.

JamesWrigley:
I think this is ready for review now, I've refactored the tests in 1791602 to be a bit cleaner.

(side note: 1791602 is the first short commit hash I've seen with only digits :four_leaf_clover:)

matheuscteo:
Thanks James! 
An idea: based on what I've seen from the ELogs, I think a nice addition would be exporting a selection of runs. Specially if we get the filtering working, I'd expect that the filter would return a selection of rows and one could export only what's selected. My suggestion would be have a context menu opened upon right clicking the selection giving the exportation option.

damnit/util.py
takluyver: Oh DAMNIT :tm: . There's a bug with how we're getting the timestamp to store in the database. You see the correct answer because the two bugs cancel each other out. Timezones! :exploding_head: 

https://github.com/European-XFEL/DAMNIT/blob/d5e5315f873bb27662b65525ca186d233e70533d/damnit/ctxsupport/ctxrunner.py#L184-L185

The Numpy datetime64 object implicitly represents time in UTC, but it doesn't store any timezone info. `.item()` converts it to a naive Python `datetime` object, which `.timestamp()` then treats as local time ([as documented](https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp)). So we're subtracting 1 or 2 hours (winter/summer) before saving it in the database.

We should be doing something like:

```python
return np.datetime64(ts, 'us').item().replace(tzinfo=timezone.utc).timestamp()
```

I'm inclined to fix this, even if it means existing databases are left showing the wrong start times in the GUI. :disappointed:

damnit/gui/main_window.py
takluyver: That's disappointing, but fair enough. :-(

damnit/util.py
takluyver: I've opened #105 to fix the extraction code.

JamesWrigley: Fixed in the GUI in 2931215, tested by checking run 290 from p4239 :clock12:

JamesWrigley:
> An idea: based on what I've seen from the ELogs, I think a nice addition would be exporting a selection of runs. Specially if we get the filtering working, I'd expect that the filter would return a selection of rows and one could export only what's selected. My suggestion would be have a context menu opened upon right clicking the selection giving the exportation option.

We could add that in the future, but I'm not sure that's what people want :thinking: My thinking is that this will mostly be used as a way to get the table data without having to work within our GUI, so I suspect users would almost certainly want all the runs in it.

PR #64, opened 2023-07-11
Make slurm jobs respect --match

JamesWrigley:
Previously the --match arguments weren't passed to the slurm job, so all variables would be reprocessed.

### DIFF
--- a/damnit/backend/extract_data.py
+++ b/damnit/backend/extract_data.py
@@ -204,6 +204,9 @@ class Extractor:
 
             python_cmd = [sys.executable, '-m', 'damnit.backend.extract_data',
                           '--cluster-job', str(proposal), str(run), run_data.value]
+            for m in match:
+                python_cmd.extend(["--match", m])
+
             res = subprocess.run([
                 'sbatch', '--parsable',
                 *self.slurm_options(),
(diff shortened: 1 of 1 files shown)


damnit/backend/extract_data.py
```diff
@@ -204,6 +204,10 @@ def extract_and_ingest(self, proposal, run, cluster=False,
 
             python_cmd = [sys.executable, '-m', 'damnit.backend.extract_data',
                           '--cluster-job', str(proposal), str(run), run_data.value]
+            if len(match) > 0:
+                for m in match:
```
tmichela, line 208: should you also check that any of the match is actually a cluster variable and only start the job in this case?

JamesWrigley: That's already done a couple lines above:
```python
        # Launch a Slurm job if there are any 'cluster' variables to evaluate
        ctx =       self.ctx_whole.filter(run_data=run_data, name_matches=match, cluster=cluster)
        ctx_slurm = self.ctx_whole.filter(run_data=run_data, name_matches=match, cluster=True)
```

This change is only to forward those match arguments to the slurm job.

tmichela: I see, thanks for the clarification. LGTM

damnit/backend/extract_data.py
```diff
@@ -204,6 +204,10 @@ def extract_and_ingest(self, proposal, run, cluster=False,
 
             python_cmd = [sys.executable, '-m', 'damnit.backend.extract_data',
                           '--cluster-job', str(proposal), str(run), run_data.value]
+            if len(match) > 0:
```
takluyver, line 207: I don't think you need this condition - if `m` is empty, the loop won't do anything anyway. :slightly_smiling_face:

takluyver:
LGTM too

damnit/backend/extract_data.py
JamesWrigley: Ah true, removed in abf6e13.

PR #65, opened 2023-07-11
Ensure that proc variables have access to raw data in the run object

JamesWrigley:
Previously proc variables executed with `ctxrunner <proposal> <run> proc` (i.e. by the listener) would also open the run with `data="proc"`, which meant that proc variables wouldn't have access to raw non-detector data.


tmichela:
LGTM!

tmichela (approved):

PR #67, opened 2023-07-12
Only attempt to change file permissions when the user is the owner

JamesWrigley:
Otherwise it will block other users from doing things like reprocessing runs or starting the listener under a different user, even if the permissions are already loose enough.

### DIFF
--- a/damnit/backend/db.py
+++ b/damnit/backend/db.py
@@ -29,7 +29,8 @@ class DamnitDB:
         log.debug("Opening database at %s", path)
         self.conn = sqlite3.connect(path, timeout=30)
         # Ensure the database is writable by everyone
-        os.chmod(path, 0o666)
+        if os.stat(path).st_uid == os.getuid():
+            os.chmod(path, 0o666)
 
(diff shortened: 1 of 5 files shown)


damnit/backend/db.py
```diff
@@ -29,7 +30,8 @@ def __init__(self, path=DB_NAME):
         log.debug("Opening database at %s", path)
         self.conn = sqlite3.connect(path, timeout=30)
         # Ensure the database is writable by everyone
-        os.chmod(path, 0o666)
+        if Path(path).owner() == getuser():
```
takluyver, line 33:
```suggestion
        if Path(path).stat().st_uid == os.getuid():
```

This should be equivalent, but it avoids some possible failure modes from translating the numeric user ID into a readable username. I'm thinking of possibilities like the user who created it no longer exists on the system, or it's created from a user inside a container. It may not be possible at all, but it's easier to avoid the lookup than to figure out if it could really come up.

JamesWrigley: Ah, good point; fixed in a938798.

PR #68, opened 2023-07-12
Add a default timeout of two minutes per test

JamesWrigley:
This is to help prevent hangs, which occasionally happens when writing tests.

I opted for a default of two minutes for each test, which is a bit more than the average time for the entire test suite in CI right now: https://github.com/European-XFEL/DAMNIT/actions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,8 @@ dependencies = [
 test = [
     "pytest",
     "pytest-qt",
-    "pytest-xvfb"
+    "pytest-xvfb",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -38,7 +38,8 @@ dependencies = [
 test = [
     "pytest",
     "pytest-qt",
-    "pytest-xvfb"
+    "pytest-xvfb",
+    "pytest-timeout"
```
takluyver, line 42:
```suggestion
    "pytest-timeout",
```

PR #73, opened 2023-07-20
Jump to run, auto-scroll and run-verticalHeader

### LINKED ISSUE #52: Fix the Run column so that it's always on the left
In databases with lots of variables it's quite annoying to scroll right in the table to a specific variable, and then scroll all the way left again to find the run number of a certain run. It'd be better if the run column was always fixed to the left side.

(I believe @matheuscteo and @CammilleCC have already implemented this but I'm leaving it open till something is merged)

### LINKED ISSUE #55: Autoscroll when new runs are added
Right now new runs are just appended to the bottom of the table and the users have to scroll down to see them, or they set the table sorting such that newer runs are at the top. We should add support for autoscrolling to the bottom anyway.

### LINKED ISSUE #57: Jump-to-run
Sometimes you know the exact run number you want to look at you want to jump straight there, we should add a button or popup for doing that (related: #36).

matheuscteo:
This PR resolves #52 , resolves #57 and resolves #55.

- A example of the jump-to-run feature can be found bellow:

Peek 2023-07-20 14-12

- Autoscroll when new runs are added can be disabled in a tableMenu action:

Peek 2023-07-20 14-21

- The frozen column with the run numbers is implemented as vertical header. I had to add a cornerButton to the table view to get a "label" for such header. This means that if one click the top-left word 'Run', the whole table is selected. I'd consider this a feature.


damnit/gui/main_window.py
```diff
@@ -437,8 +437,54 @@ def _create_menu_bar(self) -> None:
...
+        self.autoscroll_checkbox = QtWidgets.QCheckBox('Autoscroll', self)
+        self.autoscroll_checkbox.setStyleSheet("QCheckBox"
+                               "{"
+                               "padding-left : 10px;"
+                               "padding-bottom: 7px"
+                               "}")
+        self.autoscroll_checkbox.setChecked(True)        
         tableMenu = menu_bar.addMenu("Table")
+        action_autoscroll =  QtWidgets.QWidgetAction(menu_bar)
+        action_autoscroll.setDefaultWidget(self.autoscroll_checkbox)
```
takluyver, line 449: I don't think you need to fiddle around with a checkbox and a stylesheet - Qt already has a concept of a 'checkable' action. Make a regular QAction, then call `action_autoscroll.setCheckable(True)` and Qt will show it with a toggle-able checkmark.

https://doc.qt.io/qt-5/qaction.html#checkable-prop

Also, I wonder if 'autoscroll' is clear to users? Maybe something like 'Scroll to newly added runs'? :thinking:

damnit/gui/table.py
```diff
@@ -176,6 +184,34 @@ def get_static_columns(self):
...
+            styleState |= QtWidgets.QStyle.State_Sunken
+        opt.state = styleState
+        opt.rect = obj.rect()
+        # This line is the only difference to QTableCornerButton
+        opt.text = obj.text()
+        opt.position = QtWidgets.QStyleOptionHeader.OnlyOneSection
+        painter = QtWidgets.QStylePainter(obj)
+        painter.drawControl(QtWidgets.QStyle.CE_Header, opt)
+
+        return True
```
takluyver, line 212: This is... clever, but it's a lot of low-level fiddling to make the table headers look almost, but not exactly, like a regular column. 

image

I'd rather just leave the corner button blank, rely on people figuring out that the numbers are run numbers, and get rid of this code.

damnit/gui/main_window.py
matheuscteo: I agree with the label of the widged and changed to your suggestion. Yet, the checkable prop would not provide a very nice user experience, for upon clicking the action the context menu closes with this approach. I'd rather keep the QCheckBox for this reason. 

An example of this issue can be found [here](https://stackoverflow.com/questions/2050462/prevent-a-qmenu-from-closing-when-one-of-its-qaction-is-triggered).

cydanil:
Can you highlight/select the run that has been jumped to? It would make it clearer that the action happened, particularly in the second example (jumping to run 100) where it's a the bottom of the list.

matheuscteo:
> Can you highlight/select the run that has been jumped to? It would make it clearer that the action happened, particularly in the second example (jumping to run 100) where it's a the bottom of the list.

Good point. I just addressed this in [469e5b5](https://github.com/European-XFEL/DAMNIT/pull/73/commits/469e5b5b2e476e058cdc39f8803f570306271d0a). Thanks!

damnit/gui/table.py
```diff
@@ -400,6 +405,14 @@ def headerData(self, col, orientation, role=Qt.ItemDataRole.DisplayRole):
...
             name = self._data.columns[col]
             return self._main_window.column_title(name)
+        elif(
+            orientation == Qt.Orientation.Vertical
+            and role == Qt.ItemDataRole.DisplayRole
+        ):
+            row = self._data.iloc[col]['Run']
+            if pd.isna(row):
+                row = ''
+            return row
```
JamesWrigley, line 415: Is this a bugfix for something?

damnit/gui/main_window.py
takluyver: I'd prefer the menu to close when I check the box, because that's how menus usually work. I've just checked in Firefox, Pycharm, Libreoffice and Inkscape - they all have checkboxes in menus, and in all of them, clicking the checkbox closes the menu.

damnit/gui/table.py
takluyver: This is what's making the 'frozen column' - it's actually the vertical (per-row) headers. The `headerData()` method is called both for column & row headers.

takluyver:
Thanks, LGTM.

damnit/gui/table.py
JamesWrigley: Whoops thanks, I didn't read properly and thought this was in `.data()` :facepalm:

PR #76, opened 2023-07-26
Zulip support

matheuscteo:
### General information
I added some sentences for almost each function implemented in this PR. The ones I didn't mention are merely GUI helpers that do minor stuff. I'd suggest a reviewer to read through this and see the classes/functions in parallel. 
Comments in italic are more like mental notes.<sup> Feel free to read my mind. </sup> 

### ZulipMessenger
This class should be instantiated only once per opened GUI. It's only propose is to hold cache information as well as a Zulip client, which might be updated if needed. A instance of it, `zulip_messenger`, shall be always accessible from the `mainwindow` and it's created upon first request. e.g. the user tries to post a table in zulip for the first time. *So if the user doesn't want to use the integration whatsoever, nothing from the PR will be active*.

----

### Right click actions
New context menus were added in `table.py` and `plot.py`, enabling the actions `zulip_action` to be activated by the user upon right clicking the run table or a plot dialog, respectively.  These action are connected to `export_selection_to_zulip` for the `TableView` class and to `export_plot_to_zulip` for the `Canvas` class.

**export_selection_to_zulip**: Fetches selected rows. Removes "bad columns", namely  'Proposal', 'comment_id', 'Status', any column containing figures and any column that's currently invisible (`columns_invisible`, `columns_with_thumbnails`). 

Formatting the timestamp variable is also implemented. All the querying is done directly in a copy of the mainwindow.data dataframe. After the processing the `to_markdown` method is called. The resulting string is cleaned from `None` and `<NA>` values.  *Using `fillna('')` doesn't work in this dataframe, for uservars have defined types. Also, the way I see, true NaN results might be important to be spotted in the exported table*. The cleaned string is passed to the `ZulipMessenger` instance through the `send_table` method.

**export_plot_to_zulip**
Creates a tempfile and saves the figure there. Sends then the temporary path to the `ZulipMessenger` instance through the `send_plot` method.

Both `send_plot` and `send_table` are similar and execute an instance of `ZulipConfig`, a `QDialog` which has access to `ZulipMessenger`'s cache. *Dialogs are ephemeral, thus the need of creating the `ZulipMessenger` class. I've tried to add less things as possible to `mainwindow.py`, for it's already quite lengthy.*

```mermaid
%%{init: {'theme': 'green', "flowchart" : { "curve" : "basis" }} }%%
flowchart LR
 id1(TableView)  --export_selection_to_zulip--> id3(ZulipMessenger)
 id2(Canvas) --export_plot_to_zulip--> id3(ZulipMessenger)
 id3 ---- id4[send_table] & id5[send_plot]
 id5 --> id6[ZulipConfig]
 id4 --> id6[ZulipConfig]

``` 

----
### ZulipConfig
Most of the logic is within this class. Upon initiation, withing the `set_layout` method, the function `check_cache()` might be called, if `messenger`, the `ZulipMessenger`'s instance doen't contain enough information cached.

**check_cache**
Inspects the configuration file stored in `~/.local/state/damnit/.zuliprc`. The folder is the same of the table settings. If the config file is set, the parameters are passed to `messenger` and `update_client`.

**configuration file**
*The only NEEDED config parameters for the zulip client are the key and email. Yet, I added a stream and topic parameters, so we can save the last used topic and stream of the user in a way that, upon closing and reopening the GUI one can still have it cached.*

**update_client**
Tries to update/create the zulip client with the config file. Errors are logged.

**logging system and never ending try-blocks**
*I'm skeptical if all of the try-blocks I added in this class are really necessary. Still, the most important thing of this feature is that it doesn't explode anything else that's already working and since we're working with an API and many things (like dealing with list indices or dictionary keys) will depend of a successful response, that was the way I found the most cautiousness. *
Tracebacks from possible exceptions as well as API responses for which the `result` key isn't `success` (e.g. wrong API key) are added into the `output` instance of `QPlainTextEdit`. Since fetching streams and topics takes a couple seconds, a "please wait" message is also added there while the fetching is not yet done. A standard logger is also present to capture the logs.

**search_streams and _search_streams**
The logic is in `_search_streams`. `search_streams` is there only to put it all in a timeout, so the log can be logged correctly (remembering that pyqt does some async magic).
Fetching is straightforward. Both Topics and Streams are stored in the `ZulipMessenger` instance, so upon closing and reopening the dialog, searching for streams won't be needed again (unless a new stream or topic was created).

**handle_form**
Upon clicking `send` this method is called. This will call `check_for_changes`, `update_client` and `_send_msg`. 

**check_for_changes**
Checks for discrepancies between entries in the QLineEdit fields and the cached values. Returns true if there are any changes and updates the cache as well as save the new configuration with `save_config_file`.

**save_config_file**
Saves the configuration file.

**_send_msg**
Sends the message. If the `kind` is a figure, uploads the figure first. It also adds the title, if any.


damnit/gui/table.py
```diff
@@ -175,6 +178,58 @@ def get_static_columns(self):
...
+
+    def export_selection_to_zulip(self):
+        if not isinstance(self.model()._main_window.zulip_messenger, ZulipMessenger):
+            self.model()._main_window.zulip_messenger = ZulipMessenger(self.model()._main_window)       
+            
+        selected_rows = [r.row() for r in 
+                         self.selectionModel().selectedRows()]
+        bad_columns = ['Proposal', 'comment_id', 'Status']
+        df = self.model()._main_window.data.iloc[selected_rows]
+        bad_columns = bad_columns + self.columns_with_thumbnails(self.model()._data) \
```
tmichela, line 198: could you attach pictures of plots to the message if columns with thumbnails are selected?

damnit/gui/table.py
```diff
@@ -175,6 +178,58 @@ def get_static_columns(self):
...
+        self.menu.addAction(self.zulip_action)
+        self.menu.popup(QtGui.QCursor.pos())
+
+    def export_selection_to_zulip(self):
+        if not isinstance(self.model()._main_window.zulip_messenger, ZulipMessenger):
+            self.model()._main_window.zulip_messenger = ZulipMessenger(self.model()._main_window)       
+            
+        selected_rows = [r.row() for r in 
+                         self.selectionModel().selectedRows()]
+        bad_columns = ['Proposal', 'comment_id', 'Status']
```
tmichela, line 196: nitpicking, I'd not use "bad" in the name, this may be misleading... maybe something like "blacklist"(?)

damnit/gui/table.py
```diff
@@ -175,6 +178,58 @@ def get_static_columns(self):
...
+        bad_columns = list(dict.fromkeys(bad_columns))
+        df = df.drop(bad_columns, axis=1)
+        if 'Timestamp' in df.columns:
+            df['Timestamp'] = df['Timestamp'].apply(lambda dt: 
+                datetime.fromtimestamp(dt).replace(tzinfo=timezone.utc).\
+                    astimezone().strftime("%H:%M:%S %d/%m/%Y"))
+    
+        msg = df.astype(str).to_markdown(index = False)
+        # Better not to hide legit nan values
+        msg = msg.replace('None', '').replace('<NA>', '')
```
tmichela, line 209: you can replace values in the dataframe instead of the string, this is probably less prone to error, see [here](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.replace.html)

damnit/gui/zulip_messenger.py
```diff
@@ -0,0 +1,304 @@
...
+log = logging.getLogger(__name__)
+
+# This class should be instantiated only per opened GUI. It's only propose is to
+# to hold cache information as well as a Zulip client, which might be updated.
+class ZulipMessenger():
+    def __init__(self, parent = None):
+        self.main_window = parent
+        self.config_path = Path.home() / ".local" / "state" / "damnit" / ".zuliprc"
+        self.config_path.parent.mkdir(parents=True, exist_ok=True)
+        self.key, self.email, self.stream, self.topic = '','','',''
```
tmichela, line 19:
```suggestion
        self.key = self.email = self.stream = self.topic = ''
```

`str` is immutable.

damnit/gui/zulip_messenger.py
```diff
@@ -0,0 +1,304 @@
...
+        timer.timeout.connect(self._search_streams)
+        timer.start()
+        
+    def _search_streams(self):
+        changes =  self.check_for_changes(include_streams=False)
+        if changes:
+            self.update_client()
+            
+        try:
+            streams = self.messenger.client.get_streams()
```
tmichela, line 73: it's probably faster to use [get_subscriptions](https://zulip.com/api/get-subscriptions), you should then filter based on the current proposal open by DAMNIT.

damnit/gui/zulip_messenger.py
```diff
@@ -0,0 +1,304 @@
...
+        "type": "stream",
+        "to": f"{self.messenger.stream}",
+        "topic": f"{self.messenger.topic}",
+        "content": f"{self.msg}"
+        }
+        
+        try:
+            response = self.messenger.client.send_message(request)
+        except Exception as exc:
+            response = {'result' : '', 'msg': f"{exc}"}
```
tmichela, line 256: what kind of exception you expect here?

damnit/gui/zulip_messenger.py
```diff
@@ -0,0 +1,304 @@
...
+            log.info(f"{self.kind} posted to zulip stream {self.messenger.stream}, topic {self.messenger.topic}")
+            self.accept()
+        else:
+            self.show_msg(response['msg'])                
+        
+    def save_config_file(self):
+        config = ConfigParser()
+        config['api'] = {'email': self.messenger.email,
+                        'key': self.messenger.key,
+                        'site': 'https://euxfel-da.zulipchat.com',
```
tmichela, line 272: this should probably be configurable (from the environment).

pyproject.toml
```diff
@@ -31,7 +31,9 @@ dependencies = [
     "termcolor",
     "tomli",
     "pasha",
-    "extra_geom"
+    "extra_geom",
+    "zulip",
+    "tabulate"
```
tmichela, line 36:
```suggestion
```

tmichela (changes requested):

pyproject.toml
matheuscteo: Tabulate must be [installed](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_markdown.html) to use `to_markdown`. But you are right, I don't need to import it.

damnit/gui/zulip_messenger.py
CammilleCC, line 256: Looks like this is a honeypot of connection errors. The traces seem to be large and unfortunately I cannot see it further on the logs text box as it is disabled. You could either catch the known exceptions and pass a concise message or enable the text box (and set it as read-only) such that one can scroll through the error message.

CammilleCC (dismissed):
The singleton Zulip messenger approach looks good, but I am confused with the flow, especially you're calling the main window, the dialog and the singleton within each other. If you'd like, we can revisit it together some time :blush:

I am still hoping that we could drop empty columns, as these could be just noise especially with varying run types.

This is already a good start :) Would be great if you could add unit tests, but feel free to merge it this is already needed.

damnit/gui/table.py
matheuscteo: This is already supported since upon double clicking thumbnails, a Canvas instance is created. Meaning that the plot behaves like any other.

damnit/gui/plot.py
```diff
@@ -35,6 +38,9 @@ def __init__(
     ):
         super().__init__()
         self.setStyleSheet("background-color: white")
+        self.setStyleSheet("::selection{ background-color: blue }")
```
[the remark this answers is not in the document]
matheuscteo: I added this line since the line above was overwriting the background color for all elements (including the pseudo-element `selection`). A easier fix is to rewrite the line above using the QDialog selector.

matheuscteo:
### Notes on current proxy-thingy version

The current version removes the support for user defined API key and email from the GUI. Instead it uses the proxy provided by @RobertRosca . 

As far as I understood we (or someone else) would provide a API key per proposal to the instrument scientist. I idealized this to be stored in a config file, `zulip.cfg`, to be located in the same folder as the DB (i.e. one config file per experiment). An example would be the following:

```
[ZULIP]
key = usetheinstructionsrobertpostedonzuliptogetone
url =  http://exfldadev01.desy.de:8089
topics = ["topicA", "topicB", "topicC"]
```

`topics` is not a mandatory parameter. I added it there because we still must round some corners in getting the list of topic through the proxy. Meaning that, if we want a nicer user experience for the upcoming experiments, we could hardcode a list of topics there. Regardless, editing the topic manually remains supported, regardless if this parameter is provided or not.

If you want to test it locally, one could
`ssh -L 8089:exfldadev01.desy.de:8089 max-exfl-display002.desy.de`
and modify the `url` to
`url = http://localhost:8089`

I had to purge quite a bit of code, but fear not! I addressed most of the gui comments regarding the last version and they are committed  [here](https://github.com/European-XFEL/DAMNIT/commit/454be04fdda023f76294f4c5cef07c9be8cf8c47).
Meaning that if for some reason we need a version that supports user-keys, there's one.

CammilleCC (commented):
Have given a spin. Functionalities look good. Not much feedback in this iteration on my side. My opinion on the architecture still stands. Also looking forward for the tests :)

Maybe you could also check some of the code formatting, I noticed that it could do some improvements.

damnit/gui/table.py
```diff
@@ -10,9 +11,10 @@
 from PyQt5 import QtCore, QtWidgets, QtGui
 from PyQt5.QtCore import Qt
 
-
 ROW_HEIGHT = 30
 THUMBNAIL_SIZE = 35
+# The actual threshold for long messages is around 6200
+# not 10k, otherwise one gets a '414 URI Too Long' error
```
tmichela, line 17: This is strange, I'd never seen that and I've sent tons of messages over the api 🤔 
Do you see this by posting directly to zulip or through the proxy? and is it the same on zulip cloud and in our test instance?

matheuscteo: It's a proxy only issue. Directly sending with the zulip API has the usual threshold.

tmichela: OK we should check that with @RobertRosca when he's back then

RobertRosca: Had a look and the cause is likely Nginx's limit on URI length. I could bump that up, but really having the message content as a query parameter is kind of silly, so instead I'll make it a body parameter which will get past this issue.

PR #82, opened 2023-08-30
Set a default summary for the number of pulses

JamesWrigley:
This was a silly oversight, if the summary isn't specified then the variable will be displayed in the table in the `<type>: <shape>` format.


takluyver:
LGTM

PR #88, opened 2023-09-07
Swap the plotting widgets axis order so they make sense

JamesWrigley:
Now it's Y vs X instead of X vs Y, which confused a couple of people before.

(CC @daviddoji)

### DIFF
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -790,9 +790,11 @@ da-dev@xfel.eu"""
         self.plot._button_plot_runs.setMinimumWidth(200)
         plot_horizontal_layout.addStretch()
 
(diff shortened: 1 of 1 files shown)


takluyver:
Should we make this more explicit, i.e. 'X: Run, Y: Photon energy'? Just swapping them round seems likely to confuse everyone who's used to the current UI, and I don't know if everyone knows the 'Y vs X' convention (e.g. if your first language is not English).

matheuscteo:
LGTM, but I agree with @takluyver [comment](https://github.com/European-XFEL/DAMNIT/pull/88#issuecomment-1710057051)

JamesWrigley:
Good point, added in 60b7175:
image

takluyver:
LGTM

PR #89, opened 2023-09-07
Skip non-existent runs when reprocessing

JamesWrigley:
Previously these would be attempted anyway, which has some disadvantages:
- Reprocessing with e.g. `amore-proto reprocess $(seq 1 100)` would fail if any run in-between was missing, so it could take multiple invocations to reprocess a range.
- Even if a run didn't exist a database row would still be created for it, which meant that any future use of `amore-proto reprocess all` would also crash when it got to that run.

Note that this doesn't forbid including non-existent runs in the table, it just skips them when reprocessing. That's important because at some point when we hook up the DAQ as a trigger we'll record *all* created runs, not just those that have been migrated.

(CC @daviddoji)

### DIFF
--- a/damnit/cli.py
+++ b/damnit/cli.py
@@ -42,6 +42,11 @@ def excepthook(exc_type, value, tb):
     IPython.start_ipython(argv=[], display_banner=False,
                           user_ns=target_frame.f_locals | target_frame.f_globals | {"__tb": lambda: print(tb_msg)})
 
+def proposal_runs(proposal):
+    proposal_name = f"p{int(proposal):06d}"
+    raw_dir = Path(find_proposal(proposal_name)) / "raw"
+    return set(int(p.stem[1:]) for p in raw_dir.glob("*"))
+
 def main():
     ap = ArgumentParser()
     ap.add_argument('--debug', action='store_true',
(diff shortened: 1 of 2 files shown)


damnit/cli.py
```diff
@@ -185,15 +190,41 @@ def main():
...
-            for run in runs:
+
+            unavailable_runs = runs - available_runs
+            if len(unavailable_runs) > 0:
+                # Note that we print unavailable_runs as a list so it's enclosed
+                # in [] brackets, which is more recognizable than the {} braces
+                # that sets are enclosed in.
+                print(f"Warning: skipping {len(unavailable_runs)} runs because they don't exist: {list(unavailable_runs)}")
+
+            for run in (runs & available_runs):
```
takluyver, line 227:
```suggestion
                print(f"Warning: skipping {len(unavailable_runs)} runs because they don't exist: {sorted(unavailable_runs)}")

            for run in sorted(runs & available_runs):
```

Otherwise weird things can happen with the ordering:

```
>>> set(range(120, 140))
{128, 129, 130, 131, 132, 133, 134, 135, 136, 137, 138, 139, 120, 121, 122, 123, 124, 125, 126, 127}
```

damnit/cli.py
```diff
@@ -42,6 +42,11 @@ def excepthook(exc_type, value, tb):
     IPython.start_ipython(argv=[], display_banner=False,
                           user_ns=target_frame.f_locals | target_frame.f_globals | {"__tb": lambda: print(tb_msg)})
 
+def proposal_runs(proposal):
+    proposal_name = f"p{int(proposal):06d}"
+    raw_dir = Path(find_proposal(proposal_name)) / "raw"
+    return set(int(p.stem[1:]) for p in raw_dir.glob("*"))
```
takluyver, line 48: We might want to change this check in the future if we can process a run from proc or red even once the raw data is gone. But I think the concept of 'missing data for run X' makes sense, so it shouldn't be a problem to change how we check for that later. :+1:

takluyver:
Thanks, LGTM

PR #104, opened 2023-10-05
Handle all exceptions when checking for proc data

JamesWrigley:
Previously we only checked for `FileNotFoundError`'s, which would be thrown when the proc directory didn't exist at all. But sometimes the calibration pipeline will be really fast and create the proc directory and even create some files by the time the backend has received the migration message and launched a job. In this case `open_run()` will fail with an `Exception` about no HDF5 files being usable, which we didn't catch before and would cause processing of freshly-migrated runs to fail.


tmichela:
LGTM

PR #105, opened 2023-10-11
Fix getting POSIX timestamps from files

takluyver:
See the discussion on #24: Numpy gives us a naive datetime representing UTC, but then `.timestamp()` treats it as a local datetime, so we end up 1 or 2 hours out. The numeric timestamp is meant to be a number of seconds since 1st January 1970 UTC.

We have a similar bug in both the extraction side (this change) and the display change), which currently cancel each other out. So when we merge this, we should also fix the display side, which I hope we'll do in #24. And then in theory we should reprocess runs in any existing databases to get the correct timestamp.

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -182,7 +182,7 @@ def get_start_time(xd_run):
         return first_file.stat().st_mtime
     else:
         # Convert np datetime64 [ns] -> [us] -> datetime -> float  :-/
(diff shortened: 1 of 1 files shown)


JamesWrigley:
LGTM! I'd say:
- We merge this PR now
- I'll fix the GUI code in #24 
- Once both are merged I'll deploy `master` and `amore-proto reprocess all --match 'pls don't match this'` for existing databases

This is really something we should mention in a changelog too :thinking: I'll add one later.

JamesWrigley (approved):

PR #111, opened 2023-10-25
Improve support for executing a context file in another environment

JamesWrigley:
Cherry-picked at long last from one of my giant branches :sweat_smile: 

DAMNIT has had theoretical support for separate environments for a while, but semi-recently that broke. The problem is that in some places in `Extractor` and the GUI the context file needed to be loaded, and if the context file does something like `import` a package only available in the other environment, loading it will fail.

My hacky fix is to evaluate the context file in a subprocess, pickle it, and load it again in the main process. This is simple, but a bit ugly since we have to be careful to strip things that aren't pickleable (like functions) from the `ContextFile`. In the future we should do this slightly better by serializing the `ContextFile` to a JSON object or something (which might also be useful for `cammille.web`, CC @CammilleCC).

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -10,8 +10,11 @@ import functools
 import inspect
 import logging
 import os
+import pickle
+import sys
 import time
 from datetime import timezone
+import traceback
 from pathlib import Path
 from unittest.mock import MagicMock
 from graphlib import CycleError, TopologicalSorter
@@ -21,11 +24,11 @@ import h5py
 import numpy as np
 import xarray
 import sqlite3
-from scipy import ndimage
 
 from damnit_ctx import RunData, Variable, UserEditableVariable
 
 log = logging.getLogger(__name__)
+
 
 class ContextFile:
 
@@ -110,7 +113,7 @@ class ContextFile:
         code = path.read_text()
         log.debug("Loading context from %s", path)
         return ContextFile.from_str(code, external_vars)
- 
+
     @classmethod
     def from_str(cls, code: str, external_vars={}):
         d = {}
@@ -118,7 +121,7 @@ class ContextFile:
         vars = {v.name: v for v in d.values() if isinstance(v, Variable)}
         log.debug("Loaded %d variables", len(vars))
         clashing_vars = vars.keys() & external_vars.keys()
- 
+
         if len(clashing_vars) > 0:
             raise RuntimeError(f"These Variables have clashing names: {', '.join(clashing_vars)}")
         return cls(vars | external_vars, code)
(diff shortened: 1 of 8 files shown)


damnit/ctxsupport/ctxrunner.py
```diff
@@ -384,64 +409,89 @@ def train_timestamps():
...
-        proc_available = True
-    else:
-        # Otherwise check with open_run()
-        try:
-            extra_data.open_run(args.proposal, args.run, data="proc")
+    if args.subcmd == "exec":
+        # Check if we have proc data
+        proc_available = False
+        # Check if we have proc data
+        proc_available = False
```
takluyver, line 437:
```suggestion
```

This seems to have got duplicated.

damnit/ctxsupport/ctxrunner.py
```diff
@@ -384,64 +409,89 @@ def train_timestamps():
...
-    res = Results.create(ctx, inputs, args.run, args.proposal)
-
-    for path in args.save:
-        res.save_hdf5(path)
-    for path in args.save_reduced:
-        res.save_hdf5(path, reduced_only=True)
+        else:
+            # Otherwise check with open_run()
+            try:
+                extra_data.open_run(args.proposal, args.run, data="proc")
```
takluyver, line 444: Not related to this PR, but reading this code, I wonder if we can do this check only when we're going to use the result? When the raw data arrives and DAMNIT starts working on that, calibration is also starting and might have produced some files which this will try to open.

tests/test_backend.py
```diff
@@ -397,8 +398,8 @@ def slurm_scalar(run):
     from ctxrunner import main
 
     # Process run
-    with patch("ctxrunner.extra_data.open_run", return_value=mock_run):
-        main(['1234', '42', 'raw', '--save', str(out_path)])
+    with patch(f"ctxrunner.extra_data.open_run", return_value=mock_run):
```
takluyver, line 401:
```suggestion
    with patch("ctxrunner.extra_data.open_run", return_value=mock_run):
```

tests/test_backend.py
```diff
@@ -417,34 +418,82 @@ def mock_open_run(*_, data=None):
...
+    new_env_code = f"""
+    from pathlib import Path
+
+    import sfollow
+    from damnit_ctx import Variable
+
+    # Ensure that the context file is *always* evaluated in the database
+    # directory. This is necessary to ensure that things like relative imports
+    # and janky sys.path shenanigans work.
+    assert str(Path.cwd()) == "{db_dir}"
```
takluyver, line 470:
```suggestion
    assert str(Path.cwd()) == {str(db_dir)!r}
```

The `!r` should correctly manage any quotes or weird characters to produce a string literal. Not that I think there's likely to be anything weird in the path, but it's easy to do it right. :slightly_smiling_face:

takluyver (approved):
Thanks James. I've made a few minor comments inline, but this basically LGTM.

damnit/ctxsupport/ctxrunner.py
JamesWrigley: Oops :sweat_smile: Fixed in 2a041c8.

tests/test_backend.py
JamesWrigley: Fixed in 2a041c8.

tests/test_backend.py
JamesWrigley: Fixed in 2a041c8.

damnit/ctxsupport/ctxrunner.py
JamesWrigley: Yep, I ran into exactly that recently :stuck_out_tongue: #104 We could avoid it for processing non-`proc` variables, or at least when `args.run_data == 'raw'`.

takluyver (approved):
Thanks James!

PR #112, opened 2023-10-25
Fix --mock

JamesWrigley:
This was broken before (since #89) because it would always call `proposal_runs()`, which kinda defeats the purpose of `--mock`.


takluyver:
LGTM

PR #113, opened 2023-10-25
Add docs for configuring Slurm

JamesWrigley:
(CC @bermudei)


takluyver:
I might downgrade the warning to a note, but feel free to disagree. Otherwises, LGTM.

JamesWrigley:
Agreed, fixed in c521734.

PR #116, opened 2023-10-27
Convert - to _ for config keys in db-config command

takluyver:
A very simple one. I tend to forget precisely how our config keys are called, and at the command line I find it more natural to type `-` than `_`. So for the `amore-proto db-config` command, this lets you use e.g. `context-python` instead of `context_python`.

As far as I can see all our settings so far use `_` (`db_id`, `context_python`, `slurm_{partition,reservation,time}`), and I don't think we'd ever want to have two names that differ only in `_` vs. `-`.

### DIFF
--- a/damnit/cli.py
+++ b/damnit/cli.py
@@ -252,6 +252,9 @@ def main():
     elif args.subcmd == 'db-config':
         from .backend.db import DamnitDB
 
+        if args.key:
+            args.key = args.key.replace('-', '_')
+
         db = DamnitDB()
         if args.delete:
             if not args.key:


JamesWrigley (approved):
I'm a little unsure about this because it's inconsistent with `db.metameta`. Though it won't (shouldn't :flushed:) affect any users so it's probably fine :+1:

takluyver:
I guess we could move it to the `db.metameta` level if that makes more sense. It's easy enough to do, it would just have to be changed in a few more places than this variant.

PR #119, opened 2023-10-30
Add reload button to the editor

JamesWrigley:
This'll work in a pinch, but in the future I'd like to add a monitoring thread to notify the user when the context file changes on disk.
image

### DIFF
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -238,9 +238,7 @@ da-dev@xfel.eu"""
         }
         self._name_to_title = {vv : kk for kk, vv in self._title_to_name.items()}
 
-        self._editor.setText(ctx_file.code)
-        self.test_context()
-        self.mark_context_saved()
+        self.reload_context()
 
         self.extracted_data_template = str(path / "extracted_data/p{}_r{}.h5")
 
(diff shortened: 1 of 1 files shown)


damnit/gui/main_window.py
```diff
@@ -971,7 +980,7 @@ def test_context(self):
         else:
             # Move the error widget down
             height_unit = self.height() // 7
-            self._editor_parent_widget.setSizes([6 * height_unit, height_unit])
```
takluyver, line 974: Should the denominator in the line above be changed to 6? It looks like we're aiming for 1/6 (previously 1/7) the total height.

takluyver (commented):
Thanks, one very small comment, otherwise this LGTM.

I've spotted with the correlator that people click save and then reload - presumably they're understanding reload to mean something like 'load the new version into the processing pipeline'. I suspect the same thing will happen here. But I can't think of a wording that will obviously avoid that, and it's fairly harmless to click it.

damnit/gui/main_window.py
JamesWrigley: No that's the right number, the idea is that we divide the height of the window into 7 `height_unit`'s, and give 6/7 to the editor and 1/7 to the error widget (the second argument could be more clearly written as `1 * height_unit`).

takluyver: I think I made this confusing by commenting on the 'before' side of the diff rather than the 'after'. The code in this PR appears to be giving 5/7 to the editor and 1/7 to the error widget. I think Qt shares out the extra 1/7 between the widgets, though, so it doesn't matter in practice.

JamesWrigley: Ah yes I missed that, you're quite right :facepalm: I'll make a lil' fix.

PR #123, opened 2023-11-01
Fix relative heights of the editor widgets

JamesWrigley:
Spotted in https://github.com/European-XFEL/DAMNIT/pull/119#discussion_r1376379338.


JamesWrigley:
Correct, I actually have no idea why I changed that in the first place :upside_down_face: But it should be as small as possible in the case where there are no errors and doesn't need to be larger.

takluyver:
OK, then LGTM :+1:

PR #126, opened 2023-11-02
Log the --match argument during reprocessing as well

JamesWrigley:
Previously they weren't logged for variables being locally reprocessed. Also tweaked the formatting of the logging in `__main__` to be more consistent with `reprocess()`.


takluyver:
LGTM

PR #128, opened 2023-11-08
Make per-run logs accessible from the GUI

takluyver:
The actual viewer is super simple for now - just a QPlainTextEdit in read-only mode. But as log files are now per-run, they're relatively small, so I think the ability to filter them is less important. I'm sure we'll want to add features at some point, but we can start simple.

Screenshot from 2023-11-08 14-54-29

Screenshot from 2023-11-08 14-54-52

### DIFF
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -21,7 +21,7 @@ from PyQt5.QtWidgets import QMessageBox, QTabWidget, QFileDialog
 from PyQt5.Qsci import QsciScintilla, QsciLexerPython
 
 from ..backend.db import db_path, DamnitDB
-from ..backend.extract_data import get_context_file
+from ..backend.extract_data import get_context_file, process_log_path
 from ..backend import initialize_and_start_backend, backend_is_running
 from ..context import ContextFile
 from ..ctxsupport.damnit_ctx import UserEditableVariable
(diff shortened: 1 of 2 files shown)


JamesWrigley:
What happens when you try to open this 17MB log file? `/gpfs/exfel/exp/MID/202304/p003422/usr/Shared/amore/slurm_logs/r162-p3422-1295446.out`

I'm mostly just curious, no worries if it's very slow since that's a rather extreme case :grin:

takluyver:
I was expecting it to be terrible, but actually, it seemed to be perfectly fine, opening almost instantly and allowing me to scroll OK. I guess 17 MB just isn't that big. :wink:  

Screencast from 2023-11-08 16-20-54

matheuscteo:
Very nitpicking: Is there any particular reason for the log dialog to have a minimum size? 
If not, I'd suggest you to change `setMinumumSize(1000, 800)` into a simple `resize(1000, 800)`. 
The effect is the same but the user can still make the dialog smaller if necessary.

takluyver:
Good point, thanks. :+1: I think I was just looking for a `set*` method.

matheuscteo:
Thanks, LGTM.

takluyver:
Thanks! :slightly_smiling_face:

PR #133, opened 2023-11-14
Run DAMNIT on the online cluster

tmichela:
This is the current implementation to allow damnit to run on the online cluster.

### DIFF
--- a/damnit/backend/listener.py
+++ b/damnit/backend/listener.py
@@ -7,6 +7,7 @@ import queue
 import subprocess
 import sys
 from pathlib import Path
+from socket import gethostname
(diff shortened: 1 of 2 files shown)


takluyver:
LGTM

It's arguably a bit odd that we forcibly change the path even if it's explicitly passed, but in this context I think it's the right call - it needs to be hard for someone to accidentally use the offline directory.

tmichela:
Thanks, yes that's not ideal, but hopefully :crossed_fingers: only temporary.
Did not cause any issue so far as well, but I think no one is really giving explicit path and just use the proposal defaults.

PR #136, opened 2023-11-20
FIX: kafka event for offline correction

tmichela:
bug introduced in #133 the event name changed to trigger only once all detector corrections are done.


tmichela:
This was deployed on the HED branch for some time already, but we noticed it only last week as we only used the online events in the previous experiments.

PR #167, opened 2024-01-15
Add summary method (e.g. 'mean', 'sum') to database

takluyver:
In the `run_variables` table, so we know how each summary value was calculated. An empty value means no explicit summarisation - either the function returned a single value already, or we've got the default string representation of an array.

I'm working with a copy of the p900385 database. I went back to the v0 files, re-migrated to get the new `summary_method` field, and then reprocessed a run to put values in there. Here it is in sqlite-glance:

image

Part of #151


JamesWrigley (approved):
LGTM!

PR #168, opened 2024-01-15
Migration improvements

JamesWrigley:
Two changes:
- 877d05f: migrate images when doing the v0-to-v1 migration
- 68e974f: we were skipping comments :flushed:


takluyver:
So the run comments in the visible table will be handled as just an editable variable called 'comment'? My first inclination was that that feels wrong somehow, but thinking about it more, I'm coming round to it. It fits with how we're tackling other user-editable variables.

Assuming you're happy with that, this LGTM. I like the idea of folding the images migration in - I was already struggling to remember whether the order mattered, and if so which order was required.

JamesWrigley:
> So the run comments in the visible table will be handled as just an editable variable called 'comment'? My first inclination was that that feels wrong somehow, but thinking about it more, I'm coming round to it. It fits with how we're tackling other user-editable variables.

Almost, conceptually it's no different from other editable variables except that it exists by default. But *technically* it's not a user-editable variable yet because it doesn't exist in the `variables` table and there's a bunch of special-casing in the GUI to ensure that it's editable and has a specific position in the table. I wouldn't object to putting it in `variables` too, but I'll leave that for later.

takluyver:
That makes sense :+1:

PR #169, opened 2024-01-15
Migrate the `time_comments` and `variables` tables

JamesWrigley:
Previously these were just skipped.

### DIFF
--- a/damnit/migrations.py
+++ b/damnit/migrations.py
@@ -199,9 +199,23 @@ def migrate_v0_to_v1(db, db_dir, dry_run):
     new_db_path = db_dir / "runs.v1.sqlite"
     new_db_path.unlink(missing_ok=True)  # Clear any previous attempt
     new_db = DamnitDB(new_db_path)
+
+    # Copy the metadata
     for k, v in db.metameta.items():
         if k != "data_format_version":
             new_db.metameta[k] = v
+
+    # Copy the user-editable variables and standalone comments
+    for table in ["variables", "time_comments"]:
+        rows = db.conn.execute(f"SELECT * FROM {table}").fetchall()
+        if len(rows) == 0:
+            continue
+
+        placeholder = ", ".join(["?" for _ in rows[0]])
+        new_db.conn.executemany(f"""
+            INSERT INTO {table}
+            VALUES ({placeholder})
+        """, rows)
 
     # Load the data into the new database
     total_vars = 0


damnit/migrations.py
```diff
@@ -199,10 +199,25 @@ def migrate_v0_to_v1(db, db_dir, dry_run):
...
     new_db = DamnitDB(new_db_path)
+
+    # Copy the metadata
     for k, v in db.metameta.items():
         if k != "data_format_version":
             new_db.metameta[k] = v
 
+    # Copy the user-editable variables and standalone comments
+    for table in ["variables", "time_comments"]:
+        rows = [rec for rec in db.conn.execute(f"SELECT * FROM {table}")]
```
takluyver, line 210:
```suggestion
        rows = db.conn.execute(f"SELECT * FROM {table}").fetchall()
```

damnit/migrations.py
```diff
@@ -199,10 +199,25 @@ def migrate_v0_to_v1(db, db_dir, dry_run):
...
+        if len(rows) == 0:
+            continue
+
+        placeholder = ", ".join(["?" for _ in rows[0]])
+        for row in rows:
+            new_db.conn.execute(f"""
+            INSERT INTO {table}
+            VALUES ({placeholder})
+            """, row)
```
takluyver, line 219:
```suggestion
        new_db.conn.executemany(f"""
            INSERT INTO {table}
            VALUES ({placeholder})
            """, rows)
```

Doesn't really matter here, but this skips some overhead of repeatedly setting up the SQL statement.

takluyver (commented):
Thanks, good catch. :facepalm: Of course the downside of creating a whole new file is that you need to copy across anything you're not modifying. :slightly_smiling_face:

damnit/migrations.py
JamesWrigley: Added in bc7afde, though I don't think it's necessary since the cursor is iterable?

damnit/migrations.py
JamesWrigley: Ah nice, fixed in bc7afde.

damnit/migrations.py
```diff
@@ -199,10 +199,24 @@ def migrate_v0_to_v1(db, db_dir, dry_run):
...
     new_db = DamnitDB(new_db_path)
+
+    # Copy the metadata
     for k, v in db.metameta.items():
         if k != "data_format_version":
             new_db.metameta[k] = v
 
+    # Copy the user-editable variables and standalone comments
+    for table in ["variables", "time_comments"]:
+        rows = [rec for rec in db.conn.execute(f"SELECT * FROM {table}").fetchall()]
```
takluyver, line 210:
```suggestion
        rows = db.conn.execute(f"SELECT * FROM {table}").fetchall()
```

The list comprehension is redundant here - `.fetchall()` already gives you a list :wink:

takluyver:
The `.fetchall()` suggestion is basically aesthetic - I prefer to avoid unnecessary comprehensions if practical. Other than that, this LGTM.

damnit/migrations.py
JamesWrigley: Ack, I didn't read properly :facepalm: Fixed in 7d832ad.

PR #172, opened 2024-01-17
Add some (very) high-level docs for the data format

JamesWrigley:
I wanted to do this before the previous formats are lost to time. I don't think it's necessary to create a *complete* specification, just the broad strokes to give folks an idea.

This is very rudimentary and I'll probably flesh it out with some moar details in the future once v1 is properly finished.

### DIFF
--- a/docs/data-storage.md
+++ b/docs/data-storage.md
@@ -1,0 +1,108 @@
+# Data storage
+
+This page describes how DAMNIT saves the data it creates. This will probably not
+be of interest to you unless you're developing DAMNIT or writing tools that read
+data created by DAMNIT.
+
+---
+
+## Overview
+
+There are two types of storage used, both in the `usr/Shared/amore` directory of
+a proposal by default:
+
+- HDF5 files are saved in the `extracted_data/` subdirectory for each run. These
+  hold the values returned by the variable functions in the context file.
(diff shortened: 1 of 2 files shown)


docs/data-storage.md
```diff
@@ -0,0 +1,108 @@
...
+    - General DAMNIT settings for things like the slurm partition to use etc.
+
+The DAMNIT data format details the exact structure of the data in the database
+and HDF5 files.
+
+## v1 (current)
+
+In v1 there were a few changes to the way we store images and `xarray` types:
+
+- RGBA images (including thumbnails) are stored as PNG byte arrays.
```
takluyver, line 31: Only the thumbnails, but it covers monochrome images (2D arrays) as well as RGBA ones. The un-summarised data in the HDF5 files is still stored as 2D or 3D arrays.

```suggestion
- Thumbnails are stored as PNG byte arrays.
```

takluyver:
Other than that, LGTM

docs/data-storage.md
JamesWrigley: Fixed in 1db8a54. Good thing we're doing this because I thought the un-summarised data was also stored as PNGs :upside_down_face:

PR #175, opened 2024-01-19
Fix the DPI calculation for creating thumbnails

JamesWrigley:
`zoom_ratio` is always <= 1 so it should be multiplied by the `fig.dpi` to get the final DPI.

I don't think it's worth applying this fix to existing databases.


takluyver:
Thanks, good catch. I think we might need the same fix again a bit further down:

https://github.com/European-XFEL/DAMNIT/blob/939be73cf0fe9ca403661f630bce49517936de20/damnit/ctxsupport/ctxrunner.py#L375

Other than that, LGTM

JamesWrigley:
Ah yes, fixed in e901300.

PR #176, opened 2024-01-19
Add a migration for the short-lived intermediate v1 format

JamesWrigley:
Part of #151. This allows to migrate the proposals using the original v1 format to the current format. I don't intend to write tests for this migration because it's a one-off for a small number of proposals and I've already tested it on all of them :stuck_out_tongue:


damnit/migrations.py
```diff
@@ -273,3 +279,92 @@ def migrate_v0_to_v1(db, db_dir, dry_run):
...
+
+    new_db.close()
+    db.close()
+
+    if dry_run:
+        print(f"Dry-run: new format DB created at {new_db_path.name}")
+        print("If all seems OK, re-run the migration without --dry-run.")
+    else:
+        db_path = db_dir / DB_NAME
+        backup_path = db_dir / "runs.v0-backup.sqlite"
```
takluyver, line 366:
```suggestion
        backup_path = db_dir / "runs.oldfmt-backup.sqlite"
```

Not really significant, but might avoid some confusion if we need to go back to the backup for any reason, to see that this is different from v0.

takluyver:
LGTM, and I see in chat that you've been testing it, so presumably we know it works.

I wonder if we should have a subcommand a bit like `extra-data-validate` for damnit directories?

damnit/migrations.py
JamesWrigley: Good point, renamed to `runs.intermediate-v1-backup.sqlite` in 3048e2c.

JamesWrigley:
> I wonder if we should have a subcommand a bit like `extra-data-validate` for damnit directories?

Like, to check that the files have the right schema/datasets/attributes etc?

PR #177, opened 2024-01-22
Update deployment instructions

JamesWrigley:
Lets do regular installs from now on to avoid painful Python things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,5 @@ $ ssh xsoft@max-exfl.desy.de
 # Helper command to cd into the module directory and activate its environment
 $ amoremod mid
 $ git pull # Or whatever command is necessary to update the code
-
-# Only necessary if updating dependencies since DAMNIT is installed in editable mode
-$ pip install -e .
+$ pip install .
 ```


takluyver:
I'm curious if there are specific pain points you ran into with editable installs, but I can see the attraction of having the main deployment be as boring as possible, so LGTM.

JamesWrigley:
TL;DR: I screwed up the beta environment and tried to fix it by cloning the stable module environment, but that copied over a file from the editable install of the old `amore_mid_prototype` package, which led to this beauty:
```python
wrigleyj@max-exfl462 ~ % module load exfel amore/beta
wrigleyj@max-exfl462 ~ % pip show damnit
WARNING: Package(s) not found: damnit
wrigleyj@max-exfl462 ~ % python
Python 3.10.4 (main, Mar 31 2022, 08:41:55) [GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import damnit
>>> damnit.__file__
'/gpfs/exfel/sw/software/xfel_anaconda3/amore-mid/damnit/__init__.py'
```

And then @RobertRosca fixed it and said there were other potential murky issues with editable installs, so I figured it's better to stick with a regular install :upside_down_face:

PR #180, opened 2024-01-23
Add support for deleting variables

### LINKED ISSUE #69: Delete variables from the database
There is currently no way to delete variables from the database, which is very annoying if you've e.g. renamed a variable and want to delete the old one.

JamesWrigley:
This is mildly cursed because it will reload the database upon deletion. For two reasons:
- There are lingering bugs surrounding moving columns after changing the number of variables, and I don't want to deal with that right now.
- Laziness :innocent: 

My justification is that deleting a variable is something that will (hopefully) occur rarely, and occur when writing the context file for the first time while the database is likely to be empty-ish (so it won't take an obnoxious amount of time to reload). Plus, loading databases isn't *too* awful with the v1 format.

Another thing I'm unsure about is the location of the `api.delete_variable()` function. I put it in `api.py` because I have vague ideas about having a writable wrapper for a future read-only API (exposed to users through a separate package), but I couldn't think of a nice way to fit that into the `VariableData`/`RunVariables` classes. I also considered merging it with `DamnitDB.delete_variable()` but then `DamnitDB` would come to mean 'database and HDF5 files' while it's currently just 'database'. Not sure how I feel about that.

Demo:
damnit-delete-variable

Fixes #69.

### DIFF
--- a/damnit/gui/table.py
+++ b/damnit/gui/table.py
@@ -5,7 +5,9 @@ import pandas as pd
 
 from PyQt5 import QtCore, QtWidgets, QtGui
 from PyQt5.QtCore import Qt
+from PyQt5.QtWidgets import QMessageBox
 
+from ..backend.api import delete_variable
 from ..backend.db import BlobTypes
 from ..util import StatusbarStylesheet, timestamp2str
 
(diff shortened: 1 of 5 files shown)


damnit/gui/table.py
```diff
@@ -44,6 +46,10 @@ def __init__(self) -> None:
         self._columns_widget.itemChanged.connect(self.item_changed)
         self._columns_widget.model().rowsMoved.connect(self.item_moved)
 
+        self._columns_widget.setContextMenuPolicy(Qt.CustomContextMenu)
+        self._columns_widget.customContextMenuRequested.connect(self.show_delete_menu)
+        delete_action = QtWidgets.QAction("Delete")
```
tmichela, line 51:
```suggestion
```

tmichela:
Looks good enough to me, since we don't plan to maintain this ui.  
We could maybe add a comment that reloading the entire database is due to an other bug (or make an issue about it), so we can change that if the bug gets fixed.

damnit/gui/table.py
JamesWrigley: Oops :see_no_evil: Removed in 82a5e55.

JamesWrigley:
Sure, added a comment in 82a5e55.

tmichela:
LGTM

tmichela:
I just realized... what would happen if multiple clients are connected to the same database. The other clients don't know the variable is deleted and may try to access the data. Will it crash?

JamesWrigley:
Oh dear, that is an excellent point :facepalm: I'll make another PR to broadcast a deletion message.

PR #181, opened 2024-01-24
Listen for incoming events on EuXFEL Kafka broker

takluyver:
We still need to know the new topic for migration events (topics are now namespaced).

Do we want to move over the internal updates as well? We're currently using the first test broker that was set up (exflwebstor01).

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,26 +40,26 @@ at the command line.
 
 ## Kafka
 The GUI is updated by Kafka messages sent by the backend. Currently we use
-XFEL's internal Kafka broker at `exflwebstor01.desy.de:9102`, but this is only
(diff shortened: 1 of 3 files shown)


tmichela:
> Do we want to move over the internal updates as well?

Probably since `exflwgs06` seems to be the production one, the one we use currently may also disappear and or not be up for support if something happened to it.

takluyver:
Janusz M just sent me the new topic name for migrations, so I've updated that as well. This should be ready to go now.

JamesWrigley (approved):
LGTM! The readme docs will be out of date, but I'll fix that when I move them to the official docs.

takluyver:
Aha, I hadn't spotted that, thanks! I've updated it now - it's easy to do, and it seems better to keep them in sync than you having to remember to do it.

PR #182, opened 2024-01-24
Fix sending updates containing PNG thumbnails

takluyver:
There was some conversion we were doing when the reduced data was inserted into the database, which wasn't affecting the data sent through Kafka. This resulted in `<PNGData at ...>` reprs showing up in the table instead of thumbnails.

This moves the conversion to where we load the data, so the data we add to the database and the Kafka messages should be more similar. And as a bonus, it also comes out simpler.


JamesWrigley (approved):
LGTM!

PR #184, opened 2024-01-25
Bump the thumbnail size to 300x300

JamesWrigley:
This seems to fix the figure-saving errors I was seeing too :crossed_fingers:


takluyver:
LGTM

PR #186, opened 2024-01-30
Collapsible widgets

tmichela:
Some users asked to be able to hide the plotting options at the botton of the UI, as they never use it and it takes space from the table.

damnit-collapse


JamesWrigley (changes requested):
I don't know why exactly, but everything looks horizontally and vertically squished in the collapsible layout:
image

Compared to the default:
image

Not a big deal, but it would be nice if there was more padding.

JamesWrigley (approved):
LGTM!

tmichela:
thanks!

PR #190, opened 2024-01-31
Add maximize / minimize buttons to plot dialog

tmichela:
widget

### DIFF
--- a/damnit/gui/plot.py
+++ b/damnit/gui/plot.py
@@ -38,6 +38,12 @@ class Canvas(QtWidgets.QDialog):
         autoscale=True,
     ):
         super().__init__()
+        self.setWindowFlags(
+            Qt.Window |
+            Qt.WindowMinimizeButtonHint |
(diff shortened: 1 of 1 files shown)


JamesWrigley (approved):
LGTM, though they automatically show up on all the systems I've run DAMNIT on :thinking:

takluyver:
On XFCE it seems to have maximize but not minimize (the up arrow is the unusual 'roll up' option): 
image

Have you checked that minimize actually works? It's greyed out in the right click menu.

I presume this is all because it's subclassed from a dialog instead of a regular window.

tmichela:
Strange, they do not show up in the HED's control room (Ubuntu 22). Also not on my desktop (also Ubuntu 22).
But it appears on fastx indeed... :thinking:

tmichela:
> Have you checked that minimize actually works? 

yep, it works

takluyver:
LGTM, then.

PR #192, opened 2024-02-02
Refactor GUI table machinery

takluyver:
This may seem like an odd thing to do, but every time I go to work on a feature, I stall, because it seems like every part of the code knows about every other part, so it's hard to work on one piece in isolation.

This is an attempt to unpick that a bit. Specifically, I've tried to make only the `Table` class (renamed `DamnitTableModel`) know that it's actually using a pandas Dataframe to hold the data, so everything else that touches the table needs to go through that class. This isn't 100% complete yet - a couple of places in MainWindow still construct dataframes directly - but it's pretty close, and I wanted to open this before the weekend.

This shouldn't change any functionality at all; the tests are passing for me on Maxwell, and I've also tested it interactively.

### DIFF
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -28,9 +28,9 @@ from ..context import ContextFile
 from ..ctxsupport.damnit_ctx import UserEditableVariable
 from ..ctxsupport.ctxrunner import get_user_variables
 from ..definitions import UPDATE_BROKERS
-from ..util import icon_path, StatusbarStylesheet, timestamp2str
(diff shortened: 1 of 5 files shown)


tmichela:
That looks good to me assuming there's a good test coverage on the GUI, but someone with more experience with that part of the code should maybe have a look.

JamesWrigley (approved):
LGTM!

takluyver:
Thanks both! I wasn't expecting this to get reviewed so quickly, so it's a nice surprise for Monday morning. :grin:

PR #196, opened 2024-02-08
Go back to Numpy 1.26.3 for testing

takluyver:
Hopefully this fixes the segfaults we've suddenly started seeing.


takluyver:
Seems to work, will merge and see if it stays working.

PR #197, opened 2024-02-08
Fix some warnings that occur when running the tests

takluyver:
This gets rid of the warnings below - which aren't all of them, but the ones that were easiest to fix quickly.

```
tests/test_backend.py::test_extractor
  /home/kluyvert/Code/mid-2833/damnit/ctxsupport/ctxrunner.py:244:
  DeprecationWarning: The truth value of an empty array is ambiguous. Returning
  False, but in future this will result in an error. Use `array.size > 0` to check
  that an array is not empty.
    if np.isnan(ts):

tests/test_backend.py::test_extractor
  /home/kluyvert/Code/mid-2833/damnit/ctxsupport/ctxrunner.py:254:
  DeprecationWarning: parsing timezone aware datetimes is deprecated; this will
  raise an error in the future
    return np.datetime64(ts, 'us').item().replace(tzinfo=timezone.utc).timestamp()

tests/test_gui.py::test_table_and_plotting
tests/test_gui.py::test_table_and_plotting
tests/test_gui.py::test_table_and_plotting
tests/test_gui.py::test_table_and_plotting
tests/test_gui.py::test_table_and_plotting
tests/test_gui.py::test_table_and_plotting
tests/test_gui.py::test_table_and_plotting
  /home/kluyvert/Code/mid-2833/damnit/gui/plot.py:220:
  PendingDeprecationWarning: The get_cmap function will be deprecated in a future version.
  Use ``matplotlib.colormaps[name]`` or ``matplotlib.colormaps.get_cmap(obj)`` instead.
    cmap = mpl_cm.get_cmap("tab20")

tests/test_gui.py::test_table_and_plotting
tests/test_gui.py::test_table_and_plotting
  /home/kluyvert/Code/mid-2833/damnit/gui/plot.py:154:
  DeprecationWarning: the `interpolation=` argument to nanquantile was renamed to
  `method=`, which has additional options.
  Users of the modes 'nearest', 'lower', 'higher', or 'midpoint' are encouraged to review
  the method they used. (Deprecated NumPy 1.22)
    self.update_canvas(x, y, image, legend=legend)
```

### DIFF
--- a/damnit/gui/plot.py
+++ b/damnit/gui/plot.py
@@ -10,12 +10,12 @@ from PyQt5 import QtCore, QtWidgets, QtGui
 from PyQt5.QtWidgets import QMessageBox
 
 import mplcursors
+import matplotlib
 from matplotlib.backends.backend_qtagg import (
     FigureCanvas,
     NavigationToolbar2QT as NavigationToolbar,
 )
 from matplotlib.figure import Figure
-from matplotlib import cm as mpl_cm
 from mpl_pan_zoom import zoom_factory, PanManager, MouseButton
 
 from ..backend.api import RunVariables
@@ -217,7 +217,7 @@ class Canvas(QtWidgets.QDialog):
         self.figure.canvas.draw()
 
     def update_canvas(self, xs=None, ys=None, image=None, legend=None, series_names=["default"]):
-        cmap = mpl_cm.get_cmap("tab20")
+        cmap = matplotlib.colormaps["tab20"]
         self._nan_warning_label.hide()
 
         if (xs is None and ys is None) and self.plot_type == "histogram1D":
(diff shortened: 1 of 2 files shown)


damnit/gui/plot.py
```diff
@@ -279,8 +279,8 @@ def update_canvas(self, xs=None, ys=None, image=None, legend=None, series_names=
                 self._axis.set_xlabel("")
                 self._axis.set_ylabel("")
             else:
-                vmin = np.nanquantile(image, 0.01, interpolation='nearest')
-                vmax = np.nanquantile(image, 0.99, interpolation='nearest')
+                vmin = np.nanquantile(image, 0.01, method='nearest')
+                vmax = np.nanquantile(image, 0.99, method='nearest')
```
JamesWrigley, line 283: There was a weird compat issue with this that made me not fix it earlier, but now I don't remember what it was... let's do it anyway, if it breaks something we can fix it later.

JamesWrigley (approved):
LGTM!

damnit/gui/plot.py
takluyver: There's some similar code in ctxrunner which doesn't pass method/interpolation at all - maybe some target Python for running the context file had an older version of matplotlib and that bit you at some point? That's just a guess, though.

JamesWrigley: Yes, maybe that was it :thinking:

PR #201, opened 2024-02-12
Fix call to get_column_states() while exporting to Zulip

JamesWrigley:
I think this was missed in one of the recent refactorings.

CC @egorsobolev


JamesWrigley:
This one is kinda high-priority, so if there's no objections I'll merge it in the next couple of hours.

matheuscteo:
LGTM, thanks.

takluyver:
Oops, yes, I must have missed that. LGTM, thanks @JamesWrigley !

PR #202, opened 2024-02-12
Fix filtering of disabled runs

JamesWrigley:
CC @egorsobolev 

~~(I'm gonna try write a test before merging)~~

~~EDIT: on second thoughts writing a test seems a little involved so I might be lazy and leave it for another day.~~

EDIT 2: it was necessary after all :grin:


JamesWrigley:
This one is also kinda high-priority, so if there's no objections I'll merge it in the next few hours (see 21f0a62 for details).

JamesWrigley:
LMAO, the return of the segfaults.

JamesWrigley:
From trial-and-error I suspect that the explicit `win.show()` in the custom environment test might've been the issue. I removed it in 962fdf1 (the test itself should be unaffected), let's see if that helps.

PR #204, opened 2024-02-15
Fix migration of old DataArray's

JamesWrigley:
We were accidentally skipping all 1D coordinates instead of only scalar coordinates.


takluyver:
Thanks James, LGTM

PR #206, opened 2024-02-16
Fix sorting by column in the GUI

JamesWrigley:
Previously clicks on the header sections were ignored.


takluyver:
LGTM, thanks!

PR #207, opened 2024-02-16
Fix selecting rows for sending table to Zulip

takluyver:
I missed the row selection in a recent refactoring (#192), so sending the table to Zulip would always send the whole table. This should fix that.


JamesWrigley (approved):
LGTM!

PR #208, opened 2024-02-16
Update versions of Github actions in test workflow

takluyver:
Use the current version of all actions. This should fix some warnings about an old version of Node.


takluyver:
Yup, that seems to have got rid of the warnings.

PR #209, opened 2024-02-16
Add the ability to pre-create runs in the database

JamesWrigley:
This useful for setting parameters ahead of runs being taken.

Example:
damnit-precreate-runs


JamesWrigley:
Refactored it a bit in 940182e, now the bulk of the code is in `DamnitTableModel.precreate_runs()` and it doesn't bother updating `is_constant_df`.

takluyver:
This LGTM at a quick glance. I think you said you want to deploy it quickly for use at MID - if so, go ahead.

PR #214, opened 2024-02-23
Add a hidden Python console for debugging

takluyver:
This pops up a basic Python console when you press F12, to investigate objects in the GUI. It uses my newly published [Adeqt](https://pypi.org/project/adeqt/) package as an optional dependency - if that's not installed, you see an error message when trying to launch the console. I'm thinking we'll install it in the amore modules, so we can easily do a bit of debugging in a running GUI wherever it is, without requiring any special tools.

We could also make it visible in a menu, but for now I thought it might make sense to have it as a lightly hidden feature.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ cd DAMNIT
 conda create -n amore python
 
 conda activate amore
-pip install .
+pip install '.[gui]'
 ```
 
 ## Usage
@@ -81,5 +81,5 @@ $ ssh xsoft@max-exfl.desy.de
 # Helper command to cd into the module directory and activate its environment
 $ amoremod mid
 $ git pull # Or whatever command is necessary to update the code
-$ pip install .
+$ pip install '.[gui]'
 ```
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -346,6 +346,10 @@ da-dev@xfel.eu"""
         self.context_dir_changed.connect(lambda _: action_export.setEnabled(True))
         action_export.triggered.connect(self.export_table)
 
+        action_adeqt = QtWidgets.QAction("Python console", self)
+        action_adeqt.setShortcut("F12")
+        action_adeqt.triggered.connect(self.show_adeqt)
+
         action_help = QtWidgets.QAction(QtGui.QIcon("help.png"), "&Help", self)
         action_help.setShortcut("Shift+H")
         action_help.setStatusTip("Get help.")
@@ -362,6 +366,7 @@ da-dev@xfel.eu"""
         fileMenu.addAction(action_autoconfigure)
         fileMenu.addAction(action_create_var)
         fileMenu.addAction(action_export)
+        fileMenu.addAction(action_adeqt)
         fileMenu.addAction(action_help)
         fileMenu.addAction(action_exit)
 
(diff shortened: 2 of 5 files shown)


damnit/gui/main_window.py
```diff
@@ -893,6 +896,22 @@ def export_selection_to_zulip(self):
         df.replace(["None", '<NA>', 'nan'], '', inplace=True)
         self.zulip_messenger.send_table(df)
 
+    adeqt_window = None
+
+    def show_adeqt(self):
+        try:
+            from adeqt import AdeqtWindow
```
JamesWrigley, line 903: This feels like we're just making things harder for ourselves :thinking: I'd add it as a dependency.

JamesWrigley (changes requested):

damnit/gui/main_window.py
takluyver: Yeah, if we're making it visible in a menu, that definitely makes sense. :+1:

damnit/gui/main_window.py
```diff
@@ -95,6 +95,9 @@ def __init__(self, context_dir: Path = None, connect_to_kafka: bool = True):
 
         self._canvas_inspect = []
 
+        adeqt_shortcut = QtWidgets.QShortcut(QtGui.QKeySequence(Qt.Key_F12), self)
```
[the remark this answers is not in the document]
takluyver: Sure, suits me :slightly_smiling_face:

takluyver:
I've implemented both requests. :smiley: 

While I was adding a new GUI dependency, I thought I'd also take a moment to split out the GUI dependencies as an optional group, which was requested so the web code can depend on this package without pulling in PyQt stuff.

README.md
```diff
@@ -10,7 +10,7 @@ cd DAMNIT
 conda create -n amore python
 
 conda activate amore
-pip install .
+pip install .[gui]
```
JamesWrigley, line 13: This'll need quotes:
```suggestion
pip install '.[gui]'
```

(same thing below)

pyproject.toml
```diff
@@ -18,27 +18,30 @@ dependencies = [
...
 [project.optional-dependencies]
+gui = [
+    "adeqt",
+    "mplcursors",
+    "mpl-pan-zoom",
+    "openpyxl",  # for spreadsheet export
+    "PyQt5",
+    "pyflakes",  # for checking context file in editor
+    "QScintilla==2.13",
+    "tabulate",
```
JamesWrigley, line 42: Unrelated to the PR, but we can actually remove this and a couple other packages too :sweat_smile: It looks like tabulate was added when we implemented Zulip support but it was never used.

The other packages that can be removed are:
- PyYAML
- tomli
- pasha
- extra_geom
- euxfel_bunch_pattern

They were added to be used in context files before we had support for separate environments.

JamesWrigley (changes requested):

pyproject.toml
takluyver: I thought this for tabulate, but it's actually required for pandas' `to_markdown` method. I think you're right about the others, though.

README.md
takluyver: Thanks, done :+1:

takluyver:
Thanks!

PR #217, opened 2024-02-26
Delete variable in GUI without reloading entire table

takluyver:
@JamesWrigley I don't know exactly what bugs you saw before, but I tried reordering columns before & after deleting a variable, and it was working as I expected, so I hope I've fixed it. The list of columns for the popup window is regenerated each time a column is deleted.

I've followed the Qt naming convention for `removeColumns`, whereas `insert_columns` is named in Python's convention, because the remove method follows [the API Qt expects for a model class](https://doc.qt.io/qt-5/qabstractitemmodel.html#removeColumns), whereas insert does not. I know the mixture looks weird. :slightly_smiling_face:

### DIFF
--- a/damnit/gui/table.py
+++ b/damnit/gui/table.py
@@ -80,6 +80,7 @@ class TableView(QtWidgets.QTableView):
             self.model().rowsInserted.connect(self.style_comment_rows)
             self.model().rowsInserted.connect(self.resize_new_rows)
             self.model().columnsInserted.connect(self.on_columns_inserted)
+            self.model().columnsRemoved.connect(self.on_columns_removed)
             self.resizeRowsToContents()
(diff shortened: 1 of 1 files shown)


JamesWrigley (changes requested):
Maybe it was when adding columns rather than removing them, but cool :) Could you add a test for a column being removed in the GUI? (ideally at some annoying index like 0 or -1)

takluyver:
There's already a test of deleting a variable from the GUI - I've just added to it to check that the state in various bits of the GUI updates correctly. Does that look OK? It's working by name rather than by index.

JamesWrigley (approved):
Yeah that's good enough :+1: LGTM!

PR #218, opened 2024-02-27
Remove unused dependency on zulip package

takluyver:
The code to post to Zulip uses requests instead.


JamesWrigley:
Uff, slightly disturbing that we were implicitly relying on `requests` being a dependency of `zulip`...

PR #219, opened 2024-02-28
Inspect 2D summarized data

tmichela:
It is currently only possible to inspect 1d summarized data.
There's a need at HED to explore 2d data. Use case is that from run to run they have either 1 or multiple pulses per train.
In case there's 1 train the resulting data is 1d, but if there's multiple pulses it is 2d (tid vs pulse index). regardless of the case they want the summary data shown in the table and either a scatter plot (1d) or image (2d) when inspecting the variable.

the actual change is in 64f626747abc9a3fed8cb807d4e1e2bfe6e3788d else it's refactoring.

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -465,8 +465,7 @@ class Results:
 
                 if (
                     isinstance(data, (np.ndarray, xr.DataArray))
-                    and data.ndim == 1
-                    and data.shape[0] > 1
+                    and data.size > 1
                 ):
                     reduced_ds.attrs["max_diff"] = abs(
                         np.subtract(np.nanmax(data), np.nanmin(data), dtype=np.float64)
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -598,42 +598,39 @@ da-dev@xfel.eu"""
             log.warning(f"Unrecognized variable: '{quantity}'")
             return
 
-        if is_image:
-            try:
-                image = variable.ndarray()
-            except KeyError:
-                log.warning("'{}' not found in {}...".format(quantity, variable.file))
-                return
+        try:
+            data = variable.xarray()
+        except KeyError:
+            log.warning(f'"{quantity}" not found in {variable.file}...')
+            return
 
+        if is_image or data.ndim == 2:
             canvas = Canvas(
                 self,
-                image=image,
+                image=data.data,
                 title=f"{quantity_title} (run {run})",
             )
-
         else:
-            try:
-                y = variable.xarray()
-            except KeyError as e:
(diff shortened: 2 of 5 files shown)


JamesWrigley:
I'm not sure I understand, are you saying that they want e.g. a float summary but for the 2D data to be displayed as an image when inspected?

tmichela:
yes, exactly that.

damnit/gui/main_window.py
```diff
@@ -598,42 +582,39 @@ def inspect_data(self, index):
...
-            except KeyError:
-                log.warning("'{}' not found in {}...".format(quantity, variable.file))
-                return
+        try:
+            data = variable.xarray()
+        except KeyError:
+            log.warning(f'"{quantity}" not found in {variable.file}...')
+            return
 
+        if is_image or data.ndim == 2:
```
takluyver, line 591: `is_image` here is telling us whether what we display in a cell is a PNG thumbnail.  I think this is now irrelevant - we only need to check the array that we've loaded from the HDF5 file. But we should also check for 3D (RGB) arrays, in that case.

damnit/ctxsupport/ctxrunner.py
```diff
@@ -465,8 +465,7 @@ def save_hdf5(self, hdf5_path, reduced_only=False):
 
                 if (
                     isinstance(data, (np.ndarray, xr.DataArray))
-                    and data.ndim == 1
-                    and data.shape[0] > 1
+                    and data.size > 1
                 ):
```
takluyver, line 469: I would guess that arrays with 2 or more dimensions will almost always be changing, so making the summary figure bold may not be as useful. Let's go with it and see what happens - but I think it might support my case for letting the user override whether values are bold, which came up in discussing #221.

JamesWrigley:
BTW it would be nice to have a test for this: https://github.com/European-XFEL/DAMNIT/blob/master/tests/test_gui.py#L638

takluyver:
Other than that, LGTM.

tmichela:
> BTW it would be nice to have a test for this: https://github.com/European-XFEL/DAMNIT/blob/master/tests/test_gui.py#L638

I added a unit test, thanks for the pointer

damnit/gui/main_window.py
takluyver: I think those are the only cases we need. :+1:

tmichela: good there are tests :stuck_out_tongue:  , figures are converted to `rgba` so:
`if (data.ndim == 2) or (data.ndim == 3 and data.shape[-1] in (3, 4))`

tmichela:
Thanks for the review!
merging after the mandatory random segmentation fault :boom:

damnit/gui/main_window.py
takluyver: Ooh yeah, I'm glad the tests caught that.

PR #221, opened 2024-02-29
Table cell formatting with a Cell class

takluyver:
This is an idea for how the context file mechanism could expose more flexible presentation of its output. Variable functions can return a `Cell` object:

```python
from damnit_ctx import Variable, Cell
import numpy as np

@Variable(summary='sum')
def var2(run):
    return Cell(np.arange(10), bold=False)
```

Returning `Cell(x)` with no extra parameters should behave the same as returning `x` directly, but extra parameters can control formatting (this is not yet implemented in the frontend). E.g. you could colour the background according to the value. You can also provide a custom summary value to show in the cell - i.e. you're no longer limited to single Numpy functions to reduce data.

Future parameters could also influence how we present the un-summarised data when double-clicking on the cell; although that might be better set in `@Variable()`.

### DIFF
--- a/damnit/ctxsupport/damnit_ctx.py
+++ b/damnit/ctxsupport/damnit_ctx.py
@@ -91,19 +91,11 @@ class Cell:
             data = data.get_figure()
 
         if not isinstance(data, (xr.Dataset, xr.DataArray, str, type(None), Figure, PlotlyFigure)):
-            arr = np.asarray(data)
+            data = np.asarray(data)
             # Numpy will wrap any Python object, but only native arrays
             # can be saved in HDF5, not those containing Python objects.
-            if arr.dtype.hasobject:
+            if data.dtype.hasobject:
                 raise TypeError(f"Returned data type {type(data)} cannot be saved")
-            data = arr
-
-        if summary_value is not None and not isinstance(summary_value, str):
-            arr = np.asarray(summary_value)
-            if arr.dtype.hasobject:
-                raise TypeError(f"summary_value should be number or string, not {type(summary)}")
-            summary_value = arr
-
             elif not np.issubdtype(data.dtype, np.number):
                 try:
                     h5py.h5t.py_create(data.dtype, logical=True)
@@ -112,6 +104,20 @@ class Cell:
                         f"Returned data type {type(data)} whose native "
                         f"array type {data.dtype} cannot be saved",
                     )
+
+        if summary_value is not None and not isinstance(summary_value, str):
+            arr = np.asarray(summary_value)
+            if arr.dtype.hasobject:
+                raise TypeError(f"summary_value should be number or string, not {type(summary)}")
+            elif not np.issubdtype(arr.dtype, np.number):
+                try:
+                    h5py.h5t.py_create(arr.dtype, logical=True)
+                except TypeError:
+                    raise TypeError(
+                        f"Summary value {type(arr)} whose native "
+                        f"array type {arr.dtype} cannot be saved",
+                    )
+            summary_value = arr
 
         self.data = data
         self.summary = summary


takluyver:
image

CammilleCC (commented):
On first impression, it looks good! Thanks for the addition, would be happy to conveniently fetch the attributes (possible from the API?)

Would you already have ideas on how this extends on stored data display?

damnit/ctxsupport/damnit_ctx.py
```diff
@@ -69,3 +79,67 @@ def annotations(self):
...
+        d = {}
+        if self.summary is not None:
+            d['summary_method'] = self.summary
+        if self.bold is not None:
+            d['bold'] = self.bold
+        if self.background is not None:
+            d['background'] = self.background
+        if (max_diff := self._max_diff()) is not None:
+            d['max_diff'] = max_diff
+        return d
```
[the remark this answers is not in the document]
takluyver: These are already conceptually attributes for the summary, the thing we display in the table cell. We can also of course add attributes on the larger data behind that, stored in the HDF5 file, as well as the variable. Given that, namespacing them feels like overkill to me - I can't imagine that we'd want the same name for two different things. Does that seem reasonable? Or are you imagining a lot more attributes than me?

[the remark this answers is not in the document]
takluyver: I could imagine supporting data attributes in `Cell` as well, but I'd be inclined to keep that fairly flat for clarity & checkability in the context file, even if it means a lot of very specific parameters.

damnit/backend/db.py
```diff
@@ -16,11 +17,12 @@
 
 log = logging.getLogger(__name__)
 
-V1_SCHEMA = """
+V2_SCHEMA = """
```
JamesWrigley, line 20: Let's add some brief docs here: https://damnit.readthedocs.io/en/latest/internals/#data-storage

damnit/ctxsupport/damnit_ctx.py
```diff
@@ -69,3 +79,67 @@ def annotations(self):
...
+            if not re.match(r'#[0-9A-Fa-f]{6}', c):
+                raise ValueError("Colour string should be hex code (like '#ffcc00')")
+            b = bytes.fromhex(c[1:])
+            return np.frombuffer(b, dtype=np.uint8)
+        elif isinstance(c, Sequence):
+            if not len(c) == 3:
+                raise TypeError(f"Wrong number of values ({len(c)}) for R,G,B")
+            if not all(0 <= v <= 255 for v in c):
+                raise ValueError("Colour values must be 0 - 255")
+            return np.array(c, dtype=np.uint8)
```
JamesWrigley, line 116: Is there a benefit to storing it as a uint8 array? I know that's an appropriate type, but I don't think there's a technical reason in this case, and if we stored it as a `list` instead we wouldn't need to handle ndarray's specially in `get_attrs()`.

damnit/backend/extract_data.py
```diff
@@ -163,12 +163,26 @@ def get_dset_value(ds):
...
+    def get_attrs(ds):
+        d = {}
+        for name, value in ds.attrs.items():
+            if name in {"max_diff", "summary_method"}:
+                continue  # These are stored separately
+
+            if isinstance(value, np.ndarray):
+                value = value.tolist()
+            elif isinstance(value, np.generic):  # Scalar
+                value = value.item()
```
JamesWrigley, line 175: I think it'd be nice to move this logic into `Cell`.

damnit/gui/table.py
```diff
@@ -387,7 +388,7 @@ def comment_item(self, text, comment_id=None):
             item.setData(comment_id, COMMENT_ID_ROLE)
         return item
 
-    def new_item(self, value, column_id, max_diff=0):
+    def new_item(self, value, column_id, max_diff=0, attrs=None):
```
JamesWrigley, line 391: Nitpicking, does it even make sense for these arguments to have default values? I think we always want to pass them.

tests/test_backend.py
```diff
@@ -429,6 +430,31 @@ def bad(run):
         assert set(f) == {".reduced", "good", "start_time"}
         assert set(f[".reduced"]) == {"good", "start_time"}
 
+def test_results_cell(mock_run, tmp_path):
+    # Test returning an object we can't save in HDF5
```
JamesWrigley, line 434: Hmmmm :stuck_out_tongue:

tests/test_backend.py
```diff
@@ -429,6 +430,31 @@ def bad(run):
...
+    bad_obj_ctx = mkcontext(ctx_code)
+    results = bad_obj_ctx.execute(mock_run, 1000, 123, {})
+    results_hdf5_path = tmp_path / 'results.h5'
+    results.save_hdf5(results_hdf5_path)
+    with h5py.File(results_hdf5_path) as f:
+        assert f['.reduced/var2'][()] == 45
+        assert f['.reduced/var2'].attrs['bold'] == False
+        np.testing.assert_array_equal(
+            f['.reduced/var2'].attrs['background'], [0xbb, 0xdd, 0xdd]
+        )
```
JamesWrigley, line 456: It'd be good to have a test for a custom `summary_value` too.

JamesWrigley (changes requested):

damnit/ctxsupport/damnit_ctx.py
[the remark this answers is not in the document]
JamesWrigley: Whatever we decide to make the user-facing API, we can always save them differently. Though I do lean towards sticking with a flat structure.

damnit/ctxsupport/damnit_ctx.py
takluyver: Between this code (in the context process) and `get_attrs` it's stored as an HDF5 attribute, which h5py will always read as a numpy array.

damnit/backend/extract_data.py
takluyver: That doesn't work for much the same reason - the boundary between the context process (containing `Cell`) & this code is an HDF5 file, and h5py will return Numpy values.

I do want to revisit this layer as well, but I didn't want to do it as part of this PR.

damnit/gui/table.py
takluyver: You're right, we never call it without them. I think I changed this around a few times while I was working on that refactoring.

damnit/backend/db.py
takluyver: Good point, done.

tests/test_backend.py
takluyver: Thanks, good catch.

tests/test_backend.py
takluyver: It is good, adding it immediately caught a bug. :slightly_smiling_face:

takluyver:
I've rebased to clear up merge conflicts with #262.

JamesWrigley (approved):
I think this will need a rebase, but otherwise LGTM!

PR #225, opened 2024-03-12
Refactor: base GUI table on QStandardItemModel

takluyver:
This is kind of necessary as a prelude to wiring the `Cell()` class up to the table in #221, but it's also something I've been working towards for a while.

Pandas is good for working with tables of numbers & doing calculations, but it's not a great match for what we want in a GUI: it doesn't really know about formatting or images, pulling out individual values to display in the GUI is not super efficient, and adding new rows is a bit awkward. Qt provides the [QStandardItemModel](https://doc.qt.io/qt-5/qstandarditemmodel.html) to represent rich tables in memory (as opposed to lazily loaded tables), and of course it fits neatly with GUI expectations, so this is what this PR uses instead of pandas.

I also discovered a neat piece of machinery to allow sorting (QSortFilterProxyModel), which means the order within the DamnitTableModel doesn't need to change, and we can always add new rows at the end. However, this does mean the indexes we get from the view have to be translated through the proxy model to find the relevant data.

### DIFF
--- a/tests/test_gui.py
+++ b/tests/test_gui.py
@@ -52,6 +52,7 @@ def test_editor(mock_db, mock_ctx, qtbot):
 
     win = MainWindow(db_dir, False)
     win.show()
+    qtbot.addWidget(win)
     editor = win._editor
     status_bar = win._status_bar
 
@@ -140,6 +141,7 @@ def test_settings(mock_db_with_data, mock_ctx, tmp_path, monkeypatch, qtbot):
     # home directory.
     with patch("pathlib.Path.home", return_value=tmp_path):
         win = MainWindow(db_dir, False)
+    qtbot.addWidget(win)
 
     # Helper function to show the currently visible headers
     def visible_headers():
@@ -263,6 +265,7 @@ def test_handle_update(mock_db, qtbot):
     db_dir, db = mock_db
 
     win = MainWindow(db_dir, False)
+    qtbot.addWidget(win)
 
     # Helper lambdas
     model = lambda: win.table_view.model()
@@ -302,6 +305,7 @@ def test_handle_update_plots(mock_db_with_data, monkeypatch, qtbot):
     monkeypatch.chdir(db_dir)
 
     win = MainWindow(db_dir, False)
+    qtbot.addWidget(win)
     win.plot._button_plot_clicked(False)
     assert len(win.plot._canvas["canvas"]) == 1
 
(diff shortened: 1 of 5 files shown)


JamesWrigley:
(I'm dragging in Thomas M and Cammille since I'll be away for the rest of the week :stuck_out_tongue:)

takluyver:
Right, enjoy your time off :beach_umbrella:

takluyver:
@tmichela asked me about performance, so I did some simple timings with a copy of the DAMNIT folder from p6616, which has 320 runs and 6742 variables stored. I measured from running the `amore-proto gui ...` command to when the window is drawn, in FastX on max-exfl-display.

- master: 19 s, 19 s, 18.5 s
- this branch: 9.7 s, 9.3 s, 9.6 s

So at least for this case, we've got a nice speedup.

When I've profiled in the past, the `DamnitTableModel.data()` method has taken a lot of time in total. This is what Qt calls to get information from the model, and even though it's quite fast, it's called so many times that small inefficiencies mount up.

JamesWrigley:
BTW, this is quite an invasive change (and our tests are not as good as I thought they were) so I might suggest deploying it to the beta module for a couple experiments before deploying to the stable module.

takluyver:
Thanks, that's a good idea. And this should be a case where people can freely switch between the stable & beta versions, so it's easy to fall back if something breaks.

JamesWrigley (changes requested):
TBH this is too many changes to code I'm not familiar with for me to effectively review. If no one else spots anything then I'd say let's just deploy it to beta once this weeks experiments are over and see what happens :shrug:

tmichela:
> If no one else spots anything then I'd say let's just deploy it to beta once this weeks experiments are over and see what happens 

I've also looked through the code and tested it without breaking it, so I'd agree to do that. 👍

tests/test_gui.py
```diff
@@ -52,6 +52,7 @@ def test_editor(mock_db, mock_ctx, qtbot):
 
     win = MainWindow(db_dir, False)
     win.show()
+    qtbot.addWidget(win)
```
takluyver, line 55: These changes seemed to make some test failures more deterministic and easier to understand. Registering widgets like this seems to be recommended in the pytest-qt tutorial: https://pytest-qt.readthedocs.io/en/latest/tutorial.html

I'm hoping that by adding this and fixing the tests, I might have got rid of the random segfaults after the test suite finishes, but I'm not particularly confident of that.

JamesWrigley (approved):
LGTM! :crossed_fingers:

takluyver:
Thanks! Are you thinking we merge this and then deploy it to beta, or deploy it from the branch to give it some more testing before merge?

JamesWrigley:
Let's merge it and deploy to the beta, otherwise it's a bit confusing to keep track of.

tests/test_gui.py
JamesWrigley: I will miss our segfaults :pensive:

PR #226, opened 2024-03-18
Mymdc proxy

### LINKED ISSUE #34: Lightweight integration with myMdC
There are some fields in mymdc that would be useful to automatically copy over to DAMNIT, like the run type, sample type, and calibration status. The best way would probably be with a read-only token for a specific proposal that we could save in the database and use to query mymdc for each run.

(marking this as 'lightweight' integration because there are more complicated ways we could integrate with mymdc)

JamesWrigley:
In a nutshell, DAMNIT will now look for mymdc credentials in `usr/mymdc-credentials.yml` whenever a `mymdc#...` argument is used in a variable, and use them to retrieve mymdc fields (currently just the sample and run type are allowed).

The credentials file must have the format:
```yaml
token: <token>
server: <server-url>
```

Fixes #34.

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -28,6 +28,8 @@ import extra_data
 import h5py
 import numpy as np
 import xarray as xr
+import requests
+import yaml
 
 from damnit_ctx import RunData, Variable
 
@@ -43,6 +45,72 @@ class DataType(Enum):
     Dataset = "dataset"
     Image = "image"
     Timestamp = "timestamp"
+
+
+class MyMetadataClient:
+    def __init__(self, proposal, timeout=10, init_server="https://exfldadev01.desy.de/zwop"):
+        self.proposal = proposal
+        self.timeout = timeout
+        self._cache = {}
+
+        proposal_path = Path(extra_data.read_machinery.find_proposal(f"p{proposal:06d}"))
+        credentials_path = proposal_path / "usr/mymdc-credentials.yml"
+        if not credentials_path.is_file():
+            params = {
+                "proposal_no": str(proposal),
+                "kinds": "mymdc",
+                "overwrite": "false",
+                "dry_run": "false"
+            }
+            response = requests.post(f"{init_server}/api/write_tokens", params=params, timeout=timeout)
+            response.raise_for_status()
+
+        with open(credentials_path) as f:
+            document = yaml.safe_load(f)
+            self.token = document["token"]
+            self.server = document["server"]
+
+        self._headers = { "X-API-key": self.token }
+
+    def _run_info(self, run):
+        key = (run, "run_info")
+        if key not in self._cache:
+            response = requests.get(f"{self.server}/api/mymdc/proposals/by_number/{self.proposal}/runs/{run}",
+                                    headers=self._headers, timeout=self.timeout)
+            response.raise_for_status()
+            json = response.json()
+            if len(json["runs"]) == 0:
+                raise RuntimeError(f"Couldn't get run information from mymdc for p{self.proposal}, r{run}")
+
+            self._cache[key] = json["runs"][0]
+
+        return self._cache[key]
+
+    def sample_name(self, run):
+        key = (run, "sample_name")
+        if key not in self._cache:
+            run_info = self._run_info(run)
+            sample_id = run_info["sample_id"]
+            response = requests.get(f"{self.server}/api/mymdc/samples/{sample_id}",
+                                    headers=self._headers, timeout=self.timeout)
+            response.raise_for_status()
+
+            self._cache[key] = response.json()["name"]
+
+        return self._cache[key]
+
+    def run_type(self, run):
+        key = (run, "run_type")
+        if key not in self._cache:
+            run_info = self._run_info(run)
+            experiment_id = run_info["experiment_id"]
+            response = requests.get(f"{self.server}/api/mymdc/experiments/{experiment_id}",
+                                    headers=self._headers, timeout=self.timeout)
+            response.raise_for_status()
+
+            self._cache[key] = response.json()["name"]
+
+        return self._cache[key]
 
 
 class ContextFileErrors(RuntimeError):
@@ -106,6 +174,13 @@ class ContextFile:
             problems.append(
                 f"These Variables have duplicate titles between them: {', '.join(bad_variables)}"
             )
+
+        # Check that all mymdc dependencies are valid
+        for name, var in self.vars.items():
+            mymdc_args = var.arg_dependencies("mymdc#")
+            for arg_name, annotation in mymdc_args.items():
+                if annotation not in ["sample_name", "run_type"]:
+                    problems.append(f"Argument '{arg_name}' of variable '{name}' has an invalid MyMdC dependency: '{annotation}'")
 
         if problems:
             raise ContextFileErrors(problems)
@@ -191,6 +266,7 @@ class ContextFile:
 
     def execute(self, run_data, run_number, proposal, input_vars) -> 'Results':
         res = {'start_time': np.asarray(get_start_time(run_data))}
+        mymdc = None
 
         for name in self.ordered_vars():
             var = self.vars[name]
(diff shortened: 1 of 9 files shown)


JamesWrigley:
Oh, and I decided to stick with `requests` instead of `metadata_client` because that seemed like a more lightweight dependency, and the code is barely longer than using the `metadata_client` API anyway.

tmichela (changes requested):

damnit/ctxsupport/ctxrunner.py
```diff
@@ -45,6 +46,45 @@ class DataType(Enum):
...
+
+    headers={ "X-API-key": token }
+    run_res = requests.get(f"{server}/api/mymdc/proposals/by_number/{proposal}/runs/{run_no}",
+                           headers=headers, timeout=MYMDC_TIMEOUT).json()
+    if len(run_res["runs"]) == 0:
+        raise RuntimeError(f"Couldn't get run information from mymdc for p{proposal}, r{run_no}")
+
+    run_info = run_res["runs"][0]
+
+    return headers, server, run_info
```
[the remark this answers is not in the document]
takluyver: It could even be a small class, with the server URL & token as instance attributes.

damnit/ctxsupport/ctxrunner.py
```diff
@@ -45,6 +46,45 @@ class DataType(Enum):
     Timestamp = "timestamp"
 
 
+def get_run_info(run, run_no, proposal):
+    import yaml
+    import requests
```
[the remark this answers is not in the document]
takluyver: I would guess James has done this because the `ctxrunner` file is imported in DAMNIT's own frontend & backend processes, as well as the subprocesses executing the context file. By hiding the imports inside functions like this, DAMNIT doesn't require them.

We increasingly run the context file in a separate Python environment from the DAMNIT application - perhaps it's time to make that the default.

damnit/ctxsupport/ctxrunner.py
```diff
@@ -45,6 +46,45 @@ class DataType(Enum):
     Timestamp = "timestamp"
 
 
+def get_run_info(run, run_no, proposal):
```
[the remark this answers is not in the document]
JamesWrigley: Fixed in 8ce6281.

damnit/ctxsupport/ctxrunner.py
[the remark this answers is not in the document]
JamesWrigley: Yeah, though with the new class in 8ce6281 it's a little annoying to import everything where necessary so I moved them to the top.

damnit/ctxsupport/ctxrunner.py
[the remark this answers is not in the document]
JamesWrigley: And here I was trying not to make another client :smile: Added a wee one in 8ce6281.

damnit/ctxsupport/ctxrunner.py
```diff
@@ -107,6 +147,13 @@ def check(self):
                 f"These Variables have duplicate titles between them: {', '.join(bad_variables)}"
             )
 
+        # Check that all mymdc dependencies are valid
+        for name, var in self.vars.items():
+            mymdc_args = var.arg_dependencies("mymdc#")
+            for arg_name, annotation in mymdc_args.items():
+                if annotation not in ["sample", "run_type"]:
```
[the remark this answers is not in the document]
JamesWrigley: Renamed in 8ce6281.

damnit/ctxsupport/ctxrunner.py
```diff
@@ -45,6 +46,45 @@ class DataType(Enum):
...
+    proposal_path = get_proposal_path(run)
+
+    with open(proposal_path / "usr/mymdc-credentials.yml") as f:
+        document = yaml.safe_load(f)
+        token = document["token"]
+        server = document["server"]
+
+    headers={ "X-API-key": token }
+    run_res = requests.get(f"{server}/api/mymdc/proposals/by_number/{proposal}/runs/{run_no}",
+                           headers=headers, timeout=MYMDC_TIMEOUT).json()
```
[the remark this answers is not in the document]
JamesWrigley: Good point, fixed in 8ce6281.

JamesWrigley:
Oh, forgot to mention that I added support for propagating changes in the GUI in 323535e.

damnit/ctxsupport/ctxrunner.py
```diff
@@ -45,6 +47,55 @@ class DataType(Enum):
...
+
+        proposal_path = extra_data.read_machinery.find_proposal(f"p{proposal:06d}")
+        with open(proposal_path / "usr/mymdc-credentials.yml") as f:
+            document = yaml.safe_load(f)
+            self.token = document["token"]
+            self.server = document["server"]
+
+        self._headers = { "X-API-key": self.token }
+
+    def _run_info(self, run):
```
tmichela, line 64: I though this one would be the most important to cache since it's called twice for each run if you want both of sample name and run type (most likely the case).

docs/backend.md
```diff
@@ -84,6 +84,12 @@ arguments if they have the right _annotations_:
 - `meta#proposal_dir`: The root
   [Path](https://docs.python.org/3/library/pathlib.html) to the current
   proposal.
+- `mymdc#sample`: The sample from myMdC.
```
tmichela, line 87:
```suggestion
- `mymdc#sample_name`: The sample from myMdC.
```

docs/backend.md
```diff
@@ -196,6 +202,8 @@ The environment *must* have these dependencies installed for DAMNIT to work:
 
 - `extra_data`
 - `scipy`
+- `pyyaml`
+- `requests`
```
tmichela, line 206:
```suggestion
- `extra_data`
- `pyyaml`
- `requests`
- `scipy`
```

tmichela (commented):
Should we also check if the yaml file exists, and if not ask the proxy to produce one?

damnit/ctxsupport/ctxrunner.py
JamesWrigley: Added in 1d4a877.

docs/backend.md
JamesWrigley: Fixed in 1d4a877.

docs/backend.md
JamesWrigley: Fixed in 1d4a877.

JamesWrigley:
> Should we also check if the yaml file exists, and if not ask the proxy to produce one?

Excellent point, I forgot about that... added in 1d4a877.

> For the caching did you consider just using native HTTP response caching? e.g. https://requests-cache.readthedocs.io/en/stable/
> 
> Responses from MyMdC include cache info but always explicitly disable caching, spoke to ITDM about allowing cached responses and they said they'll look into it.

I didn't, but that does look like the way to go :thinking: I'd say let's stick with explicit caching for now until mymdc implements it, I wouldn't dare change the response in the proxy.

tmichela:
LGTM

@RobertRosca you said you wanted to have a closer look at that, did you have the chance to do so?

RobertRosca:
I'll have a quick look now

damnit/gui/kafka.py
```diff
@@ -38,12 +42,39 @@ def loop(self) -> None:
...
+                                   name: value
+                               }
+                           })
+
+        # Note: the send() function returns a future that we don't await. This
+        # is potentially dangerous, but the reason is because it would cause a
+        # delay in the GUI and because people don't usually close the GUI
+        # immediately after editing a variable so it's unlikely that the process
+        # will die before the update is sent (as happened with the backend).
+        self.kafka_prd.send(self.update_topic, message)
```
RobertRosca, line 60: If it's a concern, I guess one option would be to use `loop.run_soon`, or store a reference to the coroutine and, only when the gui is being closed, await it or `.close()` it?

damnit/ctxsupport/ctxrunner.py
```diff
@@ -45,6 +47,72 @@ class DataType(Enum):
     Timestamp = "timestamp"
 
 
+class MyMetadataClient:
+    def __init__(self, proposal, timeout=10, init_server="https://exfldadev01.desy.de/zwop"):
+        self.proposal = proposal
+        self.timeout = timeout
+        self._cache = {}
```
RobertRosca, line 54: With how the cache is used (keys as a tuple of arguments basically as far as I could tell?) it seems pretty similar to what using the functools `cache` decorator would do right?

e.g. 

```python
    @cache
    def _run_info(self, run):
        response = requests.get(f"{self.server}/api/mymdc/proposals/by_number/{self.proposal}/runs/{run}",
                                    headers=self._headers, timeout=self.timeout)
        ....
```

RobertRosca (approved):
Some comments but they're minor suggestions, so LGTM :+1:

tmichela (approved):

damnit/gui/kafka.py
JamesWrigley: Good idea, I added a call to `.flush()` in fbced4c which should do the same thing.

damnit/ctxsupport/ctxrunner.py
JamesWrigley: That's true, but when we make this public the caching should probably be disabled by default (at least until mymdc implements caching headers), so I'll stick with the dictionary for now so we can add an option to disable caching later.

PR #227, opened 2024-03-28
FIX: hidden header when loading a new proposal in a running GUI

tmichela:
Currently the horizontal header disappears if you load a new proposal (`SHIFT+A`):

Screenshot from 2024-03-28 15-26-23


JamesWrigley (approved):
LGTM!

PR #235, opened 2024-04-16
Clean up the API for release

### LINKED ISSUE #33: API to access the data in the table and HDF5 files
DAMNIT stores the full data from variables in HDF5 files, and reduced data and comments/user-defined variables in the SQLite database. This was always intended to be made available to users, but right now there's no API for it (albeit some users have gone to the lengths of creating their own API). See this Zulip stream for previous discussions: https://euxfel-da.zulipchat.com/#narrow/stream/323938-AMORE/topic/APIs

JamesWrigley:
Leaving this as a draft for now because it's still missing a lot of edge cases, notably around user-editable variables.

Example:
```python
damn_db = Damnit(6156)            # Create an object to refer to a database for a specific proposal
run_vars = damn_db[311]           # Select a specific run
run_vars.keys()                   # Long list of variable names
run_vars["peak_dataset"].read()   # Read one of them
```

Fixes #33.


damnit/api.py
```diff
@@ -0,0 +1,250 @@
...
+    def name(self):
+        return self._name
+
+    @property
+    def proposal(self):
+        return self._proposal
+
+    @property
+    def run(self):
+        return self._run
```
takluyver, line 45: I typically don't bother with making properties just to make attributes read-only, so e.g. `KeyData.key` and `.source` are regular attributes. I don't think we've had problems with people trying to change things like that.

Of course, there's no particular downside to making them read-only, it just makes the code a bit longer, so feel free to keep this if you prefer.

damnit/api.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def __getitem__(self, name):
+        key_locs = self._key_locations()
+        if name not in key_locs:
+            raise KeyError(f"Variable data for '{name}' not found for p{self.proposal}, r{self.run}")
```
takluyver, line 132:
```suggestion
            raise KeyError(f"Variable data for {name!r} not found for p{self.proposal}, r{self.run}")
```

This quotes the string and takes care of escaping any quotes inside it, so you can tell from the error message exactly what string got here, which could be more important now that we're allowing titles.

damnit/api.py
```diff
@@ -0,0 +1,279 @@
...
+
+        return all_keys
+
+    def keys(self):
+        return sorted(self._key_locations().keys())
+
+    def _var_titles(self):
+        result = self._db.conn.execute("SELECT name, title FROM variables").fetchall()
+        return { row[1] if row[1] is not None else row[0]: row[0] for row in result
+                 if row[0] in self.keys() }
```
takluyver, line 185: I think this calls `self.keys()` again for each key that it wants to check. That might be fast enough that it doesn't really matter, but...

```suggestion
        keys = self._key_locations()
        return { row[1] if row[1] is not None else row[0]: row[0] for row in result
                 if row[0] in keys }
```

damnit/api.py
JamesWrigley: True, but I think I'll keep them as properties so that I can document them more easily.

damnit/api.py
JamesWrigley: Fixed in d9e9f9b.

damnit/api.py
takluyver: OK :+1:

JamesWrigley:
Should we treat empty strings as being 'available' for `.keys()`? :thinking:

takluyver:
You mean, if a key is only in the database and its value there is `''` rather than `None`? Interesting one. I don't think you can easily get that state with user-editable variables, because we special case the empty string and save `None` instead:

https://github.com/European-XFEL/DAMNIT/blob/bb7ad08747c9afc4e10bae87bc960334e0fcc41f/damnit/gui/table.py#L670

You probably can for comments, though. I'd say it's probably more friendly to say you can always look up comment, even if it's empty. But I don't feel strongly about that. You could argue that treating empty comments as absent from that run is in keeping with what we do for other missing variables.

JamesWrigley:
I moved the API to the toplevel module in af7bbc9. I also tried out the include-missing-variables API but didn't really like it, it's annoying to have an extra step to check which variables are actually present :shrug:

JamesWrigley:
If there's no objections, I'd like to merge and deploy this tomorrow.

PR #238, opened 2024-04-23
Create a workflow for automatically publishing new releases

JamesWrigley:
I was going to do this manually and then I remembered we can do it automagically :stuck_out_tongue: Sneakily copied-and-pasted the code from https://github.com/European-XFEL/EXtra/pull/140.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,6 +3,7 @@ name: Tests
 on:
   push:
     branches: [master]
+    tags: ["*"]
   pull_request:
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -42,3 +43,26 @@ jobs:
...
+    needs: tests
+    permissions:
+      id-token: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
```
takluyver, line 59: Looks like there are new versions of these now:

```suggestion
        uses: actions/checkout@v4

      - name: Set up Python 3.10
        uses: actions/setup-python@v5
```

JamesWrigley: Ah good catch, fixed in 9f1ae20.

takluyver:
LGTM!

PR #239, opened 2024-04-24
Fix API dependencies

JamesWrigley:
This removes the errant dependencies the API had and adds a test for it.


JamesWrigley:
I'm roguishly going to merge this because I'd like to deploy it today :pirate_flag:

PR #247, opened 2024-05-08
Log timing after each successful variable computation

takluyver:
I basically just want to see which variables have been done in larger context files, but while we're at it we may as well record timing information as well.


JamesWrigley (approved):
LGTM!

PR #248, opened 2024-05-08
Templates for setting up new context files

takluyver:
When setting up DAMNIT for a new proposal, this gives you some options where to get the context file from:

image

Templates can have a prefix like `FXE` or `SA1`, and by default we hide those associated with another instrument or another SASE. This is based on the folder path we're going to copy it to, under `/gpfs/exfel/exp` or `/gpfs/exfel/u`. Templates which don't have one of these prefixes will be shown to all instruments.

- [x] Create a base template for SASE 3

You can also specify another proposal number to copy its context file across (but this won't copy any helper files it might need), or select a Python file manually.

In the future I hope that there might be a place where instrument groups can maintain their own templates, e.g. under `/gpfs/exfel/data/group`. But I haven't tried to implement this for now.


damnit/backend/supervisord.py
```diff
@@ -162,6 +162,7 @@ def initialize_and_start_backend(root_path, proposal=None):
         # Initialize database
         db = DamnitDB.from_dir(root_path)
         db.metameta["proposal"] = proposal
+        db.metameta["context_python"] = "/gpfs/exfel/sw/software/mambaforge/22.11/envs/202401/bin/python"
```
takluyver, line 165: I think we almost always now run the context file in our `exfel-python` environment rather than the environment where DAMNIT is installed, so that we can use tools like EXtra. So this makes that the default.

We could add more UI to let people choose the environment, but I don't want to spend time on that if there isn't a clear need.

matheuscteo:
Thanks Thomas, that's a nice addition. 

A minor possibly annoying behavior: When trying to copy the ctx file from am non-existing proposal number shows an error dialog but it closes the whole GUI just after. Ideally the user should be sent back to the form to retry without having to launch the application again. 

Apart from that LGTM

takluyver:
Good point, thanks. I've changed it so it goes back to the dialog in that case - the user can cancel the dialog if they don't want to keep trying. It's not the most elegant UI, with the dialog disappearing and reappearing, but it's also probably not something that comes up all that often, so it doesn't seem worth spending too much time on it.

PR #249, opened 2024-05-10
Preserve context file path for tracebacks

takluyver:
At present, tracebacks covering the context file don't know where the file is, so the relevant entry shows no code:

```
  File "/gpfs/exfel/sw/software/xfel_anaconda3/amore-beta/conda_env/lib/python3.10/site-packages/damnit/ctxsupport/ct
xrunner.py", line 330, in execute
    data = func(run_data)
  File "<string>", line 72, in rep_rate
  File "/gpfs/exfel/sw/software/mambaforge/22.11/envs/202401/lib/python3.10/site-packages/extra/components/pulses.py"
, line 971, in __init__
    super().__init__(data, source)
...
```

This should fix it so it can show the relevant line of code from the context file as well.

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -216,12 +216,13 @@ class ContextFile:
     def from_py_file(cls, path: Path):
         code = path.read_text()
         log.debug("Loading context from %s", path)
-        return ContextFile.from_str(code)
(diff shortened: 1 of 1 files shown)


JamesWrigley (approved):
LGTM!

takluyver:
Thanks!

Issue #240, opened 2024-04-24
Integration with `parallel`


JamesWrigley:
I regularly use `parallel` to reprocess runs in parallel, which can be quite a bit faster than doing them serially. It would be nice if there was a `-j<n>` option for `amore-proto reprocess` that would create the right `parallel` command and run it for the user.

JamesWrigley:
Closing this because now we support using slurm for everything.

PR #250, opened 2024-05-10
Add option to reprocess runs in parallel

takluyver:
To address #240.

This isn't particularly elegant code, but it does work, at least in a quick test. The most awkward bit is with the extractor object: I can't just use the same one across all threads, because at least the sqlite database connection shouldn't be shared across threads. But I also don't want to create a new Extractor for each individual run, so I didn't just `map()` over the available runs. I'm still thinking about how to handle this better.

I wanted to show less output when processing in parallel, because the usual output would probably be too fast and too jumbled up to be much use. The longer output is still directed to the relevant file in the `process_logs` folder.

### DIFF
--- a/damnit/backend/extract_data.py
+++ b/damnit/backend/extract_data.py
@@ -4,7 +4,6 @@ import getpass
 import os
 import logging
 import pickle
-import queue
 import re
 import shlex
 import socket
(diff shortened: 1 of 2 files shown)


JamesWrigley:
For the record, I'm fine with creating extractors for each run :) That's essentially what's happening now when I use `parallel` anyway.

takluyver:
Thanks, that's good to know. I vaguely remember that sometimes setting up the Kafka producer could take a while, so for recomputing one small variable in several runs it was an annoying overhead to recreate it each time.

I'll have a look (next week) if I can find a neater way to do it, otherwise I can fall back to just creating a new Extractor for each run.

takluyver:
I'm going to close this, because I think #270 meets the need better, by submitting jobs to the solaris cluster to run in parallel.

PR #255, opened 2024-05-27
mount remote filesystem with sshfs

tmichela:
Attempt to solve executing the context file with online data without having 2 separate databases.

The current solution is very simple: it mounts the remote directory with sshfs and the execution runs on maxwell.
I think ideally, when we have a centralized backend, we may want to have a permanent  NFS mount but that's for a longer term solution.
There's a meta variable to set in order to run the context on online data.
I've loosely tested it and I couldn't see a noticeable slowdown due to the remote data in the kind of data we work with (at HED).

Some limitations:
- one has to (purposely) start the backend on one of the display nodes with the 10G interface to the online cluster (to avoid causing issues going through the gateway machine)
- we skip cluster variables. This would complicate things as regular cluster nodes do not have direct connection to the online cluster. And realistically, cluster variable are expected to take longer time so I don't think there would be any requirement to run those on online data.


damnit/ctxsupport/ctxrunner.py
```diff
@@ -594,6 +598,89 @@ def save_hdf5(self, hdf5_path, reduced_only=False):
...
+@contextmanager
+def filesystem(host='localhost'):
+    """Mount a remote filesystem with sshfs"""
+    if host == 'localhost':
+        yield
+        return
+
+    with TemporaryDirectory() as td:
+        try:
+            run(f'sshfs {host}:{ed.read_machinery.DATA_ROOT_DIR} {td}', check=True, shell=True)
```
JamesWrigley, line 611: Will this ever prompt the user for credentials?

JamesWrigley: Ok, sounds good :+1: I dunno if it's possible, but is there a way to get `sshfs` to exit with an error if it can't authenticate without a prompt? That would make debugging a bit easier if we ever get into that situation.

tmichela: yea we can pass ssh options, so we can deactivate password authentication:

```shell
sshfs ... -o ssh_command='ssh -o PasswordAuthentication=no'
```

damnit/backend/extract_data.py
```diff
@@ -374,6 +382,7 @@ def reprocess(runs, proposal=None, match=(), mock=False):
     ap.add_argument('run_data', choices=('raw', 'proc', 'all'))
     ap.add_argument('--cluster-job', action="store_true")
     ap.add_argument('--match', action="append", default=[])
+    ap.add_argument('--data-location', default='localhost')
```
takluyver, line 385: Do we want to make this a hidden option? I assume that usually by the time you're reprocessing something, data will be on Maxwell anyway, and I'm not sure we want to make this too obvious, in particular if it's likely to go away in the future.

If so, this should hide it from the `--help` output:

```suggestion
    ap.add_argument('--data-location', default='localhost', help=argparse.SUPPRESS)
```

damnit/backend/listener.py
```diff
@@ -73,20 +78,30 @@ def __init__(self, context_dir=Path('.')):
...
+        # check backend host and connection to online cluster
+        self.online_data_host = None
+        self.run_online = self.db.metameta.get('run_online', False) is not False
+        if self.run_online and gethostname() not in BACKEND_HOSTS_TO_ONLINE:
+            log.warning(f"Disabled online processing, the backend must run on one of: {BACKEND_HOSTS_TO_ONLINE}")
+            self.run_online = False
+        if self.run_online:
+            topic = Path(find_proposal(f'p{self.proposal:06}')).parts[-3]
+            if (remote_host := ONLINE_HOSTS.get(topic)) is None:
+                log.warn(f"Can't run online processing for topic '{topic}'")
```
takluyver, line 98:
```suggestion
                log.warning(f"Can't run online processing for topic '{topic}'")
```

The log 'warn' alias is deprecated, although it will probably never actually be removed.

damnit/ctxsupport/ctxrunner.py
```diff
@@ -594,6 +598,98 @@ def save_hdf5(self, hdf5_path, reduced_only=False):
...
+    with TemporaryDirectory() as td:
+        try:
+            mount_command = [
+                "sshfs", f"{host}:{extra_data.read_machinery.DATA_ROOT_DIR}", str(td),
+                # deactivate password prompt to fail if we don't have a valid ssh key
+                "-o", "ssh_command='ssh -o PasswordAuthentication=no'"
+            ]
+            run(mount_command, check=True, shell=True)
+
+            with patch("extra_data.read_machinery.DATA_ROOT_DIR", td):
```
takluyver, line 623: Should we make this a parameter for `open_run()` or have some other way of passing it in? It's OK for now, but this is relying on implementation details of extra_data.

damnit/backend/listener.py
```diff
@@ -73,20 +78,30 @@ def __init__(self, context_dir=Path('.')):
         # Fail fast if read-only - https://stackoverflow.com/a/44707371/434217
         self.db.conn.execute("pragma user_version=0;")
         self.proposal = self.db.metameta['proposal']
-        log.info(f"Will watch for events from proposal {self.proposal}")
 
-        if gethostname().startswith('exflonc'):
-            # running on the online cluster
-            kafka_conf = KAFKA_CONF['onc']
-        else:
-            kafka_conf = KAFKA_CONF['maxwell']
```
takluyver, line 82: Should this be replaced with some check that stops people running the listener on the ONC if they're used to doing that? Or are all such listeners started by you?

takluyver:
I think this generally LGTM.

As I understand it, it will listen for all 4 possible events at all times. The online events will only do anything if we set a flag in the database. Then it will process any non-cluster variables with the data mounted through sshfs. Once the data is migrated to Maxwell (and possibly corrected), the same variables will be reevaluated with data in the usual location, plus any `cluster=True` variables will run. So there's a bit of duplicated effort, but this should be used with relatively small data and light computation, so it shouldn't matter too much. Does that all sound correct?

damnit/backend/extract_data.py
tmichela: Thanks! Yes, I meant to hide that argument, we don't want to reprocess data on the online cluster

damnit/backend/listener.py
tmichela: Good point. People at the instrument know how and do start the backend regularly.

tmichela:
Yep, you got it right. I know there's some duplication, we could e.g. ignore offline processing (at least for raw data) if online is set. But I think for the use case we have this should not cause issues.  
I also want to keep this simple as this is likely a temporary solution. We can always reiterate after some useage.

There are also other ideas in mind for long term, for example Philipp suggested that file based migration might be there *soon* so that could be an alternative to remove this special case. If we can also special case the calibration pipeline to run as soon as the data is available on maxwell. But that's an other conversation

tmichela:
I also prevented now the GUI to start the backend if it's running on the online cluster.

takluyver:
Other than that, LGTM

damnit/gui/main_window.py
```diff
@@ -929,6 +929,14 @@ def __init__(self, file_path: Path, parent=None):
 def prompt_setup_db_and_backend(context_dir: Path, prop_no=None, parent=None):
     if not db_path(context_dir).is_file():
 
+        if gethostname().endswith('exflonc'):
```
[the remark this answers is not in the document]
tmichela: ouch 🫣

damnit/ctxsupport/ctxrunner.py
tmichela: I agree if we use this in the long run, but considering this is only a temporary solution (look like we may get slurm on the online cluster :crossed_fingers: ) I'd say it's better to not add extra arguments to extra_data

tmichela:
I'll test this a bit more, and will merge end of this week or early next week, unless there are objections.

tmichela:
main change since the last review: it's adapted to work running on the solaris cluster, tunneling data mounts through the display nodes

damnit/backend/extraction_control.py
```diff
@@ -85,12 +85,14 @@ class ExtractionRequest:
     match: tuple = ()
     mock: bool = False
     update_vars: bool = True
+    data_location: str = 'localhost'
```
takluyver, line 88: I see what you're doing, but I think either the name `data_location` or the default 'localhost' is rather confusing. The data location is always on a network filesystem in any case, the distinction is whether it's persistently mounted at a standard path or whether we mount it temporarily at an unpredictable one.

tmichela: yea, I did not think too hard about names (I thought this would disappear soon ™️  anyway). I'm open for suggestions, as I'm not sure what would make it clearer. `mount='default'`?

takluyver: Yeah, `mount` or `mount_host` might be clearer. I might also make the default `None` to make it clear that this is actually different behaviour rather than just a different target.

PR #256, opened 2024-05-29
Evaluate non-cluster variables in Slurm using Solaris cluster

takluyver:
Instead of running extraction as a subprocess of the listener on the login node (or wherever the listener was started), this would run it in the [Solaris cluster](https://confluence.desy.de/display/MXW/Running-single-core-batch-job_302030316.html), which consists of relatively low-powered nodes configured to allow multiple small jobs at once. The `cluster=True` variables are still run on a dedicated node in the main cluster, as before.

The default settings seem to give jobs 2 CPU cores and 4 GB of RAM. We can specify different limits if we want - the nodes seem to have 20 physical cores (40 with hyperthreading) and 256 GB RAM. I'm still figuring out how the different options interact (e.g. `--mem` and `--mem-per-cpu`).

Going over the specified memory limit can get the job killed, whereas on a login node we might get away with using more RAM for a while, but cause problems unpredictably. In the long run, I think having a consistent rule about what variables need `cluster=True` is probably preferable, but it might be frustrating while we adapt.

I've only done this for the listener at the moment - reprocessing still runs on whatever node you run `amore-proto reprocess` on. If we do want to go in this direction, I'll make that use the solaris cluster too.

### DIFF
--- a/damnit/backend/listener.py
+++ b/damnit/backend/listener.py
@@ -4,6 +4,7 @@ import logging
 import os
 import platform
 import queue
+import shlex
 import subprocess
 import sys
 from pathlib import Path
@@ -145,13 +146,24 @@ class EventProcessor:
         log.info("Processing output will be written to %s",
                  log_path.relative_to(self.context_dir.absolute()))
 
-        with log_path.open('ab') as logf:
-            # Create subprocess to process the run
-            extract_proc = subprocess.Popen([
-                sys.executable, '-m', 'damnit.backend.extract_data',
-                str(proposal), str(run), run_data.value
-            ], cwd=self.context_dir, stdout=logf, stderr=subprocess.STDOUT)
-        self.extract_procs_queue.put((proposal, run, extract_proc))
+        python_cmd = [
+            sys.executable, '-m', 'damnit.backend.extract_data',
+            str(proposal), str(run), run_data.value
+        ]
+
+        res = subprocess.run([
+            'sbatch', '--parsable',
+            '--cluster=solaris',
+            '-o', log_path,
+            '--open-mode=append',
+            # Note: we put the run number first so that it's visible in
+            # squeue's default 11-character column for the JobName.
+            '--job-name', f"r{run}-p{proposal}-damnit",
+            '--wrap', shlex.join(python_cmd)
+        ], stdout=subprocess.PIPE, text=True)
+        job_id = res.stdout.partition(';')[0].strip()
+        log.info("Launched Slurm (solaris) job %s to run context file", job_id)
+
 
 def listen():
     # Set up logging to a file
--- a/damnit/backend/listener.py
+++ b/damnit/backend/listener.py
@@ -115,6 +115,8 @@ class EventProcessor:
             'sbatch', '--parsable',
             '--cluster=solaris',
             '-o', log_path,
+            '--cpus-per-task', '4',
+            '--mem', '25G',
             '--open-mode=append',
             # Note: we put the run number first so that it's visible in
             # squeue's default 11-character column for the JobName.
--- a/damnit/backend/listener.py
+++ b/damnit/backend/listener.py
@@ -3,13 +3,11 @@ import json
 import logging
 import os
 import platform
-import queue
 import shlex
 import subprocess
 import sys
 from pathlib import Path
 from socket import gethostname
-from threading import Thread
 
 from kafka import KafkaConsumer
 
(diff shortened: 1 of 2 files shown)


JamesWrigley (changes requested):
I would suggest these limits (that I mostly made up):
- 8 cores (sometimes people [*cough* me] will do a small amount of parallel processing on raw data)
- 50GB of memory (this feels like the most data one could easily load and work with in memory without having to wait too long)

Those limits would allow us at most 5 (per-node) * 12 (nodes in the cluster) = 60 concurrent jobs (assuming no one else is using the cluster), which feels like plenty.

And my preference would be to only do this in the listener, for two reasons:
- When reprocessing manually for testing it's much more convenient to have the job run locally and see the logs immediately.
- We're mostly trying to protect against being thundering herd-ed, which is something that can only happen through the listener (unless the user tries very hard with `parallel`).

The downside is that jobs that succeed locally may fail on the cluster because of a lack of resources. I'm ok with that tradeoff because I think it's unlikely to occur very often, but I don't feel very strongly about it.

takluyver:
Those limits don't sound unreasonable, but I might be inclined to go a bit smaller by default - maybe half that, so 1/10th of a node. We could have some config options to increase the limits per proposal - and if we use that often, increasing the defaults is easy. Decreasing the defaults, e.g. if these nodes get substantially busier, may well be trickier, if things that people are used to doing suddenly don't work.

What do you think?

JamesWrigley:
Sure, that sounds fine :+1:

JamesWrigley:
Is this ready to be reviewed/merged? The only things I'd miss are some options to specify different default limits.

takluyver:
I think it's ready, so long as we're happy with the overall idea. I'll add the settings now, it should be very simple.

takluyver:
I've called the options `noncluster_cpus` and `noncluster_mem` for now. I don't love 'noncluster' as a name, if you have a better way to describe these.

damnit/backend/listener.py
```diff
@@ -145,13 +106,27 @@ def handle_event(self, record, msg: dict, run_data: RunData):
...
+            str(proposal), str(run), run_data.value
+        ]
+
+        res = subprocess.run([
+            'sbatch', '--parsable',
+            '--cluster=solaris',
+            '-o', log_path,
+            # Default 4 CPU cores & 25 GB memory, can be overridden
+            '--cpus-per-task', str(self.db.metameta.get('noncluster_cpus', '4')),
+            '--mem', self.db.metameta.get('noncluster_mem', '25G'),
```
JamesWrigley, line 120: These options should be in the documentation. Also not a fan of the names, but I can't think of anything better so let's roll with it :shrug:

JamesWrigley (changes requested):

damnit/backend/listener.py
takluyver: Thanks, I've described them (along with the default limits) in the docs.

docs/backend.md
```diff
@@ -111,6 +111,19 @@ def baz(run, value: "var#foo"=42):
     return value
 ```
 
+Variable functions can use up to 4 CPU cores and 25 GB of RAM by default.
```
JamesWrigley, line 114: This is slightly misleading because it implies that those limits are always set, whereas they're only used for automatically launched jobs from the listener. I'd reword this to something like:
```suggestion
To avoid accidentally running out of resources when processing runs automatically (i.e. triggered by migration or calibration during an experiment), the backend will launch its jobs for *non-cluster* variables with limits of 4 CPU cores and 25 GB of RAM by default. This only affects jobs launched automatically by the backend, running `amore-proto reprocess` manually or specifying `cluster=True` will still allow using the entire node.
```

docs/backend.md
```diff
@@ -111,6 +111,19 @@ def baz(run, value: "var#foo"=42):
...
+If more resources are needed, use `cluster=True` (described below) to access all
+of the cores & memory of an assigned cluster node. If required, you can also
+change the limits for non-cluster variables :
```
JamesWrigley, line 117: Nitpicking:
```suggestion
change the limits for non-cluster variables:
```

JamesWrigley (changes requested):

docs/backend.md
takluyver: I was avoiding this topic for now in the interests of getting this merged, but...

I actually think that reprocessing should work by submitting a Slurm job too, and I'd like to work on that soon, but I think it's a bit more complicated, so I didn't want to hold this bit up. Having consistent rules for non-cluster variables seems like a good thing - you're less likely to write something that works when you test it and then fails on new runs, and I think it will make triggering reprocessing from the GUI (Qt or web) easier as well.  Plus it's less to explain. :slightly_smiling_face: 

I think we can figure something out to show the processing output 'live' in the terminal for `amore-proto reprocess`, either using Slurm's own `srun` command, or something like our [sfollow](https://github.com/European-XFEL/sfollow) wrapper.

JamesWrigley (approved):
LGTM!

takluyver:
Thanks!

PR #257, opened 2024-05-30
Improve error handling when opening variables in the GUI

JamesWrigley:
Two changes:
- Handle the case where the HDF5 file isn't found. This may happen if there are multiple proposals in the same database, then `RunVariables` will attempt to look for an HDF5 file for the currently configured proposal and not be able to find it. We should fix this properly by adding support to the API for databases with multiple proposals.
- Show a status message if there was an error instead of just logging something to the terminal.


tmichela:
LGTM

PR #262, opened 2024-06-10
Save plotly figure

tmichela:
Allow saving plotly figures as a variable result.
GUI widget will be as a separate PR.

### DIFF
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -87,6 +87,14 @@ class VariableData:
         """
         return self._h5_path
 
+    def type_hint(self):
+        """Type hint for this variable data.
+
+        one of ``DataType`` or None.
+        """
+        with self._open_h5_group() as group:
+            return self._type_hint(group)
+
     @contextmanager
     def _open_h5_group(self):
         with h5py.File(self._h5_path) as f:
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -1,16 +1,16 @@
 import os
-from glob import iglob
 import os.path as osp
-from enum import Enum
-from pathlib import Path
 from contextlib import contextmanager
+from enum import Enum
+from glob import iglob
+from pathlib import Path
 
 import h5py
 import pandas as pd
 import plotly.io as pio
 import xarray as xr
 
-from .backend.db import DamnitDB, BlobTypes
+from .backend.db import BlobTypes, DamnitDB
 
 
 # This is a copy of damnit.ctxsupport.ctxrunner.DataType, purely so that we can
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -129,8 +129,9 @@ class VariableData:
 
             dset = group["data"]
             if type_hint is DataType.PlotlyFigure:
-                # json string representing a plotly Figure
-                return pio.from_json(dset[()])
+                # plotly figures are json serialized and saved as uint8 arrays
+                # to enable compression in HDF5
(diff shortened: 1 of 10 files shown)


damnit/ctxsupport/ctxrunner.py
```diff
@@ -14,24 +14,23 @@
...
 import numpy as np
-import xarray as xr
 import requests
+import xarray as xr
 import yaml
-
 from damnit_ctx import RunData, Variable
+from matplotlib.axes import Axes
+from matplotlib.figure import Figure
+from plotly.graph_objects import Figure as PlotlyFigure
```
JamesWrigley, line 33: We need to document the new dependency: https://damnit.readthedocs.io/en/latest/backend/#using-custom-environments

JamesWrigley (changes requested):
Should we do this automagically for all Figure's? I'm inclined to say yes, but let's see how it works first :octopus: Also, how large are the datasets that are produced? I imagine they can get into multiple megabytes pretty easily, is it worth compressing the JSON first?

damnit/ctxsupport/ctxrunner.py
takluyver: Alternatively, we could do the check without making plotly a dependency, using a version of the `isinstance_no_import` function we've got in a couple of places already:

https://github.com/European-XFEL/EXtra/blob/7d0f66f781321f84b73790f27ccba11854a7608a/src/extra/components/utils.py#L30-L36

(I got this technique from IPython)

tmichela:
> Should we do this automagically for all Figure's?

What do you mean by "all Figure"? 🤔 

> Also, how large are the datasets that are saved?  I imagine they can get into multiple megabytes pretty easily, is it worth compressing the JSON first?

About 3MB for XGM data example. A quick test gives me about 50% compression on that data.

JamesWrigley:
> What do you mean by "all Figure"? 🤔

As in, converting regular matplotlib figures to plotly figures by default. I'm assuming that's possible from the XGM example? :eyes: 

> About 3MB for XGM data example. A quick test gives me about 50% compression on that data (using gzip).

Ok, yeah I think I'd go for that then.

damnit/api.py
```diff
@@ -19,6 +20,7 @@ class DataType(Enum):
     Dataset = "dataset"
     Image = "image"
     Timestamp = "timestamp"
+    PlotlyFigure = "PlotlyFigure"
```
CammilleCC, line 23: Nitpick: `PlotlyJson = "plotly_json` perhaps?

CammilleCC (commented):
LGTM, thanks for the addition!

tmichela:
> As in, converting regular matplotlib figures to plotly figures by default. I'm assuming that's possible from the XGM example? 👀

huh, no I don't think that's possible. I just re-created the plots for comparison.

JamesWrigley:
Ahh, devastating...

damnit/api.py
tmichela: I prefer `PlotlyFigure` as this is the object that is to be saved, and returned by the API.

tmichela:
> Ok, yeah I think I'd go for that then.

Compression is implemented. Not sure it's super worth it but 🤷

JamesWrigley (approved):
LGTM!

tmichela:
Thanks!

PR #263, opened 2024-06-10
Show tooltip with full size thumbnail for image items

tmichela:
Because it's hard to quickly figure out what complex figure are on cells.

image

### DIFF
--- a/damnit/gui/table.py
+++ b/damnit/gui/table.py
@@ -1,14 +1,15 @@
 import logging
 import time
+from base64 import b64encode
 from itertools import groupby
 
-from PyQt5 import QtCore, QtWidgets, QtGui
+from PyQt5 import QtCore, QtGui, QtWidgets
 from PyQt5.QtCore import Qt
 from PyQt5.QtWidgets import QMessageBox
 
-from ..backend.db import ReducedData, BlobTypes, DamnitDB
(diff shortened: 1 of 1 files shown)


damnit/gui/table.py
```diff
@@ -377,6 +378,9 @@ def text_item(self, value, display=None):
     def image_item(self, png_data: bytes):
         item = self.itemPrototype().clone()
         item.setData(self.generateThumbnail(png_data), role=Qt.DecorationRole)
+        item.setToolTip(
+            f'<img src="data:image/png;base64,{b64encode(png_data).decode()}">'
```
JamesWrigley, line 382: Am I correct in thinking that we would need to specify `width` and `height` to control the size of the tooltip?

tmichela: I assume all thumbnails are saved as 300x300 px, right?
but yea, it does not hurt to enforte it here

JamesWrigley: Yes, I'm just asking out of curiosity in case we want to change it in the future. I think leaving them at full resolution is fine though, it looks decently sized when displayed.

JamesWrigley (approved):
LGTM!

tmichela:
Thanks!

PR #264, opened 2024-06-10
GUI: Plotly widget

tmichela:
follow up for #262

### DIFF
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -87,7 +87,6 @@ class VariableData:
         """
         return self._h5_path
 
-    @property
     def type_hint(self):
         """Type hint for this variable data.
 
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -87,6 +87,15 @@ class VariableData:
         """
         return self._h5_path
 
+    @property
+    def type_hint(self):
+        """Type hint for this variable data.
+
+        one of ``DataType`` or None.
+        """
+        with self._open_h5_group() as group:
+            return self._type_hint(group)
+
     @contextmanager
     def _open_h5_group(self):
         with h5py.File(self._h5_path) as f:
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -583,7 +583,7 @@ da-dev@xfel.eu"""
                                      stylesheet=StatusbarStylesheet.ERROR)
             return
 
-        if variable.type_hint is DataType.PlotlyFigure:
+        if variable.type_hint() is DataType.PlotlyFigure:
             pp = PlotlyPlot(variable)
             self._canvas_inspect.append(pp)
             pp.show()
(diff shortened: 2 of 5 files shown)


damnit/api.py
```diff
@@ -87,6 +87,15 @@ def file(self) -> Path:
         """
         return self._h5_path
 
+    @property
+    def type_hint(self):
```
JamesWrigley, line 91: I'm inclined to make this a method because opening the file might be expensive.

damnit/gui/main_window.py
```diff
@@ -983,6 +992,12 @@ def run_app(context_dir, connect_to_kafka=True):
             # User said no to setting up a new database
             return 0
 
+    # configure webviewer url engine
+    global SCHEME_HANDLER, WEB_PROFILE
```
JamesWrigley, line 996: Does `SCHEME_HANDLER` need to be global? And can't we pass `WEB_PROFILE` to `PlotlyPlot()` directly?

JamesWrigley (changes requested):

damnit/gui/main_window.py
tmichela: right, fixed

JamesWrigley (approved):
I'm a little nervous about depending on `PyQtWebEngine` (packaging can be a little horrendous), but c'est la vie. LGTM!

damnit/gui/web_viewer.py
```diff
@@ -0,0 +1,87 @@
+from PyQt5 import QtCore, QtWidgets
```
CammilleCC, line 1: Maybe clean up the imports? For instance, `QMovie` is not used.

damnit/gui/web_viewer.py
```diff
@@ -0,0 +1,87 @@
...
+
+class PlotlyPlot(QtWidgets.QWidget):
+    def __init__(self, variable):
+        super().__init__()
+        self.setGeometry(0, 0, 1024, 768)
+
+        self.url = f"damn.it{variable._db.path.parent}/{variable.proposal}/{variable.run}/{variable.name}"
+
+        self.browser = QWebEngineView(self)
+        profile = QWebEngineProfile.defaultProfile()
```
CammilleCC, line 61: Curious question: what's the difference of the profile that was declared on the `main_window` and the profile that was set here? Would it be possible to reuse it?

CammilleCC (commented):
Looks good, haven't had the chance to test it locally but the logic seems sound. It's fascinating that one has to do the hacky way to implement a URI scheme just to show large HTML

damnit/gui/web_viewer.py
tmichela: they are the same. Here I get the [default profile](https://doc.qt.io/qt-6/qwebengineprofile.html#defaultProfile), I don't set a new one.

PR #266, opened 2024-06-12
Explicitly specify maxwell cluster when submitting larger jobs

takluyver:
This will hopefully fix the 'invalid partition specified' errors


takluyver:
I'm deploying this as a hotfix now

JamesWrigley (approved):
LGTM!

takluyver:
Thanks! I'm going to keep an eye on FXE's p5436 until another run is migrated and check that this is behaving as expected before I merge it.

takluyver:
Yup, that seems to be working

PR #271, opened 2024-06-20
Prevent variable results with dtypes that cannot be converted to HDF5

philsmt:
SQS tried to directly return the result of `DataCollection.train_timestamps()` from a variable, which is a dtype not convertable to HDF5 datatype IDs, thus failing when the dataset is written. The current code only prevents `object` arrays, this PR adds a check for convertability for anything that is not an `np.number`.

As a side note, throwing an unchecked exception in the writing code also created some hard to read stack traces full of red herings.

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -345,6 +345,16 @@ class ContextFile:
                             name, type(data)
                         )
                         data = None
+                    elif not np.issubdtype(arr.dtype, np.number):
+                        try:
+                            h5py.h5t.py_create(arr.dtype, logical=True)
+                        except TypeError:
+                            log.error(
+                                "Variable %s returned %s whose native "
+                                "array type %s cannot be saved",
+                                name, type(data), arr.dtype
+                            )
+                            data = None
                     else:
                         data = arr
             except Exception:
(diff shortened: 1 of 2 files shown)


JamesWrigley (approved):
LGTM!

damnit/ctxsupport/ctxrunner.py
```diff
@@ -345,6 +345,15 @@ def execute(self, run_data, run_number, proposal, input_vars) -> 'Results':
                             name, type(data)
                         )
                         data = None
+                    elif not np.issubdtype(arr.dtype, np.number):
+                        try:
+                            h5py.h5t.py_create(arr.dtype, logical=True)
+                        except TypeError:
+                            log.error("Variable %s returned %s which cannot "
+                                      "be converted to an HDF5 datatype object",
+                                      name, type(data)
```
takluyver, line 354: I think the error message here should mention the dtype of the array, because the type will usually be something like ndarray, or maybe list, so it could be confusing to say that can't be converted to HDF5.

philsmt: Good point. I changed the message to include both types, making clear the dtype may be the result of some form of conversion. For non-collection types the actual type may be more useful, while for collection types listing both may make it more likely to make the connection.

takluyver:
Thanks, LGTM

PR #272, opened 2024-06-21
FIX: renaming data vars in xarray DataSet objects before saving to HDF5

tmichela:


JamesWrigley (approved):
LGTM! I'm surprised we didn't hit this before...

PR #273, opened 2024-06-24
HDF5 array compression

tmichela:
Extend compression in HDF5 to numpy arrays, xarray DataArray and DataSet.

A very quick test on random runs gives me about 30% size reduction.


JamesWrigley (approved):
LGTM! Though I think @takluyver should take a squiz at it too. I assume the read overhead will be negligible for the vast majority of arrays.

damnit/ctxsupport/ctxrunner.py
```diff
@@ -611,15 +611,26 @@ def save_hdf5(self, hdf5_path, reduced_only=False):
...
-
-            obj.to_netcdf(hdf5_path, mode="a", format="NETCDF4", group=name, engine="h5netcdf")
+                vars_names = {}
+                for var_name, dataarray in obj.items():
+                    if var_name is not None and "/" in var_name:
+                        vars_names[var_name] = var_name.replace("/", "_")
+                    dataarray.encoding.update(COMPRESSION_OPTS)
+                obj = obj.rename_vars(vars_names)
+
+            obj.to_netcdf(
```
takluyver, line 627: [`to_netcdf()`](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.to_netcdf.html#xarray.DataArray.to_netcdf) takes an `encoding=` parameter. Does this work? If it does, it seems neater than modifying the objects to change how they're stored.

damnit/ctxsupport/ctxrunner.py
```diff
@@ -36,7 +36,7 @@
 log = logging.getLogger(__name__)
 
 THUMBNAIL_SIZE = 300 # px
-
+COMPRESSION_OPTS = {'compression': 'gzip', 'compression_opts': 1, 'fletcher32': True, 'shuffle': True}
```
takluyver, line 39: Is there a specific reason to use the fletcher32 checksum? We dropped it from the gain/mask datasets in corrected data back in 2021, because we wanted to bypass HDF5 for writing, and we were assured that the filesystem is doing checksums already, so there's no particular need to do them at the HDF5 chunk level.

takluyver:
> I assume the read overhead will be negligible for the vast majority of arrays.

I hope so, although gzip is relatively slow both on compression & decompression by modern standards. But hopefully the benefits of reading less data balance it out.

There is a potential drawback of enabling compression on arrays of unknown shape, however. Compression in HDF5 requires chunking, and if you don't specify a chunk shape, h5py will guess it, aiming to make it equally good for slicing along any axis. But that can easily mean equally *bad*, especially if the chunks in one slice exceed the chunk cache size (a paltry 1 MB by default), leaving you repeatedly decompressing the same data.

Where we know the nature of the data, it's usually easy to manually specify a chunk shape to make access efficient. But in this case we're trying to handle whatever the context file returns. :shrug:

damnit/ctxsupport/ctxrunner.py
tmichela: >  we were assured that the filesystem is doing checksums already

Then I'll remove that option. I just remembered that it was used in the calibration code.

tmichela:
My feeling is that a variable result is most likely a single entity one would want to load at once to use. And that's what our API does anyway at the moment, so yea 🤷 
Do you think it can be so bad that we should not try to compress on arbitrary arrays? the 30% reduction is pretty appealing though.

takluyver:
I think that's probably reasonable enough, at least to start with and see how we go. Results have to be able to fit in memory when they're created, because of the way context files work.

It wouldn't entirely surprise me if someone decides to store large-ish results (10s of GB) this way and wants to load parts of them, either with an extension to our API or using the files directly. But we can cross that bridge when we come to it. :+1:

damnit/ctxsupport/ctxrunner.py
tmichela: if *should* work but you still need to map variable names to encoding args, i.e. `{'var1': {...}, 'var2': {...}, ...}`.  
Another caveat I've read is that it overwrites any potential existing information the dataarray encoding.

takluyver: OK, fair enough. Though that does make me think, should we apply the changes in a way that leaves existing values if they're there? Or does xarray create the array with default options?

tmichela: do you mean, e.g. not overwrite the compression algorithm or level if it is already defined?  
I think I'm more enclined to overwrite these always (at least the compression algorithm) 🤔

tmichela: I believe `encoding` is an empty dict by default, but I imagine we could use it in the `Variable` code do improve compression on a variable basis with chunking, quantization, ...

takluyver: > do you mean, e.g. not overwrite the compression algorithm or level if it is already defined?

That was what I was thinking, e.g. if you knew data was likely to be incompressible, you could tell it not to try by setting `arr.encoding['compression'] = None`. I don't feel strongly about this, though, so if you're happy overwriting them, go for it.

tmichela: I'd be fine with that, just worried (although I sense this is unlikely to be an issue) that someone tries to use a non standard compression filter, or one that isn't supported in our environment.

takluyver: Your call - I'm also happy to let it try compressing everything the same way and add more options later if the need arises. :slightly_smiling_face:

takluyver:
LGTM!

damnit/ctxsupport/ctxrunner.py
tmichela: I changed my mind and keep original encoding options from the data array.

PR #276, opened 2024-06-28
Fix returning e.g. bool from variable functions

takluyver:
Just ran into this, I'm going to deploy it as a hotfix


JamesWrigley (approved):
Would be nice to get a test in, but otherwise LGTM!

takluyver:
Thanks! I'll add a test on Monday. I've deployed this on both the `mid` & `beta` modules.

PR #277, opened 2024-07-02
Add a database config option to skip starting a listener

takluyver:
This skips the 'backend not running' dialog in the GUI, and the listener also checks for this after 10 minutes with no messages, to shut itself down.

I see two use cases for this:

1. Working with old proposals where you know no new runs are coming, so there's no point running a listener.
2. As part of migration from per-proposal listeners to one central listener, set this for proposals where the central listener will process new runs.


tmichela:
LGTM,

I was also thinking about maybe getting the proposal information from mymdc and decide from the status to start the backend or not. But since we want/will move to a single listener model, this does not matter much.

takluyver:
Yeah, it would be neat to decide automatically using myMdC, but I don't think it's worth the extra complexity. It's not a big deal if a listener is left running after the beamtime, in any case.

Thanks for the review!

PR #279, opened 2024-07-03
Simplify testing CLI by passing argv in

takluyver:
This lets the `main()` function takes an optional list of command line args, simplifying testing because you don't need to keep modifying `sys.argv` I've used this pattern on various other projects before.


tmichela:
LGTM

takluyver:
Thanks :+1:

PR #280, opened 2024-07-04
Defer imports for context file runner code

takluyver:
This uses the `isinstance_no_import` trick, avoiding loading modules just to check if objects belong to their types.

This reduces the list of packages required for the context Python environment. It can also reduce the overhead of launching the context runner process, although it still loads everything that the context file imports, so the change probably isn't dramatic for most real scenarios.


tmichela:
LGTM

takluyver:
Thanks again :smile:

PR #281, opened 2024-07-04
Defer some imports in damnit.api

takluyver:
Loading anything from damnit loads the api module (so that you can do `from damnit import Damnit`), but some of these imports can be noticeably slow. Deferring these 3 to where they're used brings `amore-proto db-config` down from 0.6 to 0.3 seconds with a warm cache. The times are worse with a cold cache, but it's hard to get useful measurements for that.

There are other ways we could get a similar result: we could make `damnit.Damnit` lazy-load (I happened to see plotly does a lot of lazy-loading), or we could make separate packages. This way is the smallest change.

### DIFF
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -6,9 +6,6 @@ from glob import iglob
 from pathlib import Path
 
 import h5py
-import pandas as pd
-import plotly.io as pio
-import xarray as xr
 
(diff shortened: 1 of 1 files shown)


CammilleCC:
LGTM

takluyver:
Thanks Cammille!

PR #283, opened 2024-07-05
Add support for Python 3.12

JamesWrigley:
See the commit messages for details. I don't think we need to switch to 3.12 now, but I'm using it locally so it's nice to have it working.

Also ran into this fun bug: https://github.com/matplotlib/matplotlib/issues/28235
Which required creating an environment with: `conda create -c conda-forge -n amore python expat==2.6.1`


takluyver:
LGTM!

PR #285, opened 2024-07-10
(Re)process runs from the GUI

### LINKED ISSUE #39: Allow the users to reprocess runs from the GUI
Reprocessing runs is one of the few times we force users to use a terminal, which is scary for lots of them. It should be possible to reprocess runs and specific variables from the GUI.

takluyver:
Another attempt at this, now that processing happens in Slurm.

image

Closes #39

### DIFF
--- a/damnit/gui/process.py
+++ b/damnit/gui/process.py
@@ -1,0 +1,207 @@
+import logging
+import re
+
+from PyQt5 import QtWidgets
+from PyQt5.QtCore import Qt
+from PyQt5.QtWidgets import QDialogButtonBox
+
+from ..context import RunData
+from ..backend.extraction_control import ExtractionRequest
+
+log = logging.getLogger(__name__)
+
+run_range_re = re.compile(r"(\d+)(-\d+)?$")
+
+RUNS_MSG = "Enter run numbers & ranges e.g. '17, 20-32'"
+
+deselected_vars = set()
+
+
+def parse_run_ranges(ranges: str) -> list[int]:
+    res = []
+    for run_s in ranges.split(","):
+        if m := run_range_re.match(run_s.strip()):
+            start = int(m[1])
+            if m[2] is None:
+                end = start
+            else:
+                end = int(m[2][1:])
+                if start > end:
+                    return []
+            res.extend(range(start, end + 1))
+        else:
+            return []
+
+    return res
+
+
+def fmt_run_ranges(run_nums: list[int]) -> str:
+    if not run_nums:
+        return ""
+
+    range_starts, range_ends = [run_nums[0]], []
+    current_range_end = run_nums[0]
+    for r in run_nums[1:]:
+        if r > current_range_end + 1:
+            range_ends.append(current_range_end)
+            range_starts.append(r)
+        current_range_end = r
+    range_ends.append(current_range_end)
+
+    s_pieces = []
+    for start, end in zip(range_starts, range_ends):
+        if start == end:
+            s_pieces.append(str(start))
+        else:
+            s_pieces.append(f"{start}-{end}")
+
+    return ", ".join(s_pieces)
+
+
+class ProcessingDialog(QtWidgets.QDialog):
+    selected_runs = ()
+    all_vars_selected = False
+    no_vars_selected = False
+
+    def __init__(self, proposal: str, runs: list[int], var_ids_titles, parent=None):
+        super().__init__(parent)
+
+        self.setWindowTitle("Process runs")
+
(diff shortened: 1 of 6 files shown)


tmichela:
Could you also add an option on right click?

tmichela:
it'd also be nice to check if new variables exist in the context file that were never processed, otherwise it will always be skipped, unless one selects all variables.

damnit/gui/process.py
```diff
@@ -0,0 +1,207 @@
...
+        main_vbox.addLayout(hbox1)
+
+        grid1 = QtWidgets.QGridLayout()
+        hbox1.addLayout(grid1)
+        vbox2 = QtWidgets.QVBoxLayout()
+        hbox1.addLayout(vbox2)
+
+        self.edit_prop = QtWidgets.QLineEdit(proposal)
+        #self.edit_prop.setInputMask('999900;_')  # 4-6 digits
+        grid1.addWidget(QtWidgets.QLabel("Proposal:"), 0, 0)
```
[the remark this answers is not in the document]
takluyver: I'm trying to preserve the possibility of one DAMNIT table covering multiple proposals, without getting in the way of the common case - the proposal number should be pre-filled if possible, and the run field gets the focus. Do you prefer to push people back to the CLI if multiple proposals ever comes up.

takluyver:
I'll add a right-click option. :+1: 

For new variables: would it be a decent compromise to update the variables whenever you successfully save the context file in the GUI? Loading the context file often takes a few seconds, so doing it when showing this dialog either means an awkward pause before showing it, or a janky update after the list shows up, neither of which seem like great UX.

I'm also wondering about an option to process only variables which don't already have results, to fill in newly added variables.

damnit/gui/table.py
```diff
@@ -60,6 +60,8 @@ def __init__(self) -> None:
         self.show_logs_action = QtWidgets.QAction('View processing logs')
         self.show_logs_action.triggered.connect(self.show_run_logs)
         self.context_menu.addAction(self.show_logs_action)
+        self.process_action = QtWidget.QAction('Reprocess runs')
```
tmichela, line 63:
```suggestion
        self.process_action = QtWidgets.QAction('Reprocess runs')
```

tmichela:
> For new variables: would it be a decent compromise to update the variables whenever you successfully save the context file in the GUI? 

It is really that long to validate the file? 🤔  
What do you think of having a "load new variables" button then on the widget? or a "process new variable" checkbox?

damnit/gui/process.py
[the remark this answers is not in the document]
tmichela: I honestly hope we won't have to support this case 😅 but sure, why not...

tmichela:
I'm thinking that I would be nice to have a visual indicator of the runs processing, particularly in this case of processing from the gui. But I think we don't have a way to monitor slurm jobs status yet, right?

takluyver:
> It is really that long to validate the file? 🤔

It's now common/normal to use a different Python environment for evaluating the context file from running the DAMNIT application. So validating it means launching a subprocess and then loading all the modules it imports. Between GPFS, which isn't great for lots of small files, and xarray, which gets its `__version__` in a convenient but inefficient way, it can easily be a couple of seconds. Not a long time, but definitely not instant for GUI code, hence #291.

> I'm thinking that I would be nice to have a visual indicator of the runs processing

Yup, I'm also planning that, but I wanted to do this first.

damnit/gui/process.py
```diff
@@ -0,0 +1,207 @@
...
+            self.vars_list.addItem(itm)
+
+        self.vars_list.itemChanged.connect(self.validate_vars)
+
+        self.validate_runs()
+        self.validate_vars()
+
+        self.edit_runs.setFocus()
+
+    def validate_runs(self):
```
tmichela, line 127: I would probably be useful to cehck if the selected runs exist, and:
- show a warning message if they don't exist
- do not start a job for runs that do not exist
- show an error if trying to reprocess and none of the selected runs exist

takluyver: OK, how about like this...

1. All chosen runs valid:

image

2. Some don't exist:

image

3. No runs exist, or the proposal number is wrong, or the user running the GUI doesn't have access:

image

In the last case, the 'Ok' button is disabled so you have to change it to launch reprocessing.

tmichela: Great!

tmichela:
I'm happy with the current state, feel free to merge.
We can address the other points later if they are needed.

LGTM

takluyver:
Thanks for the review!

PR #287, opened 2024-07-11
Add back --direct option to run reprocessing directly on local node

takluyver:
I may have been overly pushy about always doing processing in Slurm. Apart from anything else, running locally with the `--mock` option is useful for testing without being on Maxwell.

So this adds an `amore-proto reprocess --direct` option to run processing in subprocesses. Like `--watch` this will process runs one at a time, with live output in the terminal (as well as written to the log file).

cc @tmichela


tmichela:
Thanks! LGTM

PR #289, opened 2024-07-15
Plot xarray

tmichela:
Special case `xarray.DataArray` in the plot canvas, to retain coordinate and labels information from the data.

### DIFF
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -592,17 +592,29 @@ da-dev@xfel.eu"""
             pp.show()
             return
 
+        if variable.type_hint() is DataType.Dataset:
+            QMessageBox.warning(self, "Can't inspect variable",
+                                f"'{quantity}' is a Xarray Dataset (not supported).")
+
         try:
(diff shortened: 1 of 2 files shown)


damnit/gui/main_window.py
```diff
@@ -592,17 +592,29 @@ def inspect_data(self, index):
...
+                                f"'{quantity}' is a Xarray Dataset (not supported).")
+
         try:
-            data = xr.DataArray(variable.read())
+            data = variable.read()
         except KeyError:
             log.warning(f'"{quantity}" not found in {variable.file}...')
             return
 
+        if variable.type_hint() is DataType.DataArray:
```
takluyver, line 605: If DataArrays are always handled by this branch, I think we can simplify the code below a bit. It creates a DataArray object and checks for a trainId coordinate, which it will now never find. So we could probably just make that code work with a numpy array.

takluyver:
Other than that (and it's not an important point), LGTM.

tmichela:
Thanks

PR #290, opened 2024-07-15
Save QAction's as class variables

JamesWrigley:
This fixes segfaults that I came across locally. Presumably the actions were getting GC'd before the slots that use them were signalled.

### DIFF
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -324,22 +324,22 @@ da-dev@xfel.eu"""
         )
         action_autoconfigure.triggered.connect(self._menu_bar_autoconfigure)
 
-        action_create_var = QtWidgets.QAction(
+        self.action_create_var = QtWidgets.QAction(
             QtGui.QIcon.fromTheme("accessories-text-editor"),
             "&Create user variable",
             self
         )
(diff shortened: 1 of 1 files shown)


tmichela:
LGTM

tmichela (approved):

JamesWrigley:
And apparently also introduces new segfaults...

JamesWrigley:
Going off the logs I'd say it's some kind of race condition involving plotly/kaleido? Kaleido at least is what's appearing in the logs. So far it's consistently failing at `test_table_and_plotting` but that shouldn't touch any plotly stuff. Not really sure where the extra thread is coming from either.

Anyway, I think this PR is still an improvement so I'll merge it. I don't think it's the cause of the bug because the same failure happened once on #284.

takluyver:
Thanks! I always get confused about when Qt does & doesn't keep objects alive.

> Not really sure where the extra thread is coming from either.

Looks like kaleido starts a thread internally as part of managing its subprocess: https://github.com/plotly/Kaleido/blob/056642fe8994354838a7cb619f6a98a7ad0608ab/repos/kaleido/py/kaleido/scopes/base.py#L185-L189

PR #291, opened 2024-07-19
Check context file without blocking GUI thread

takluyver:
I noticed some time ago that a large part of what makes starting the GUI slow is waiting for the context file check, which usually involves starting a subprocess with a different Python, and often also doing some fairly heavy imports in the context file.

This pushes these operations into a separate thread, so we can start using the GUI before the context file check finishes. It also means saving the context file doesn't block the GUI.

Naturally, more async makes the tests a bit more complicated, but I think it's manageable.

### DIFF
--- a/damnit/gui/editor.py
+++ b/damnit/gui/editor.py
@@ -4,8 +4,8 @@ from io import StringIO
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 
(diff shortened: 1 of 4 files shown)


tmichela:
LGTM

takluyver:
Oops, I missed the LGTM. Thanks!

PR #301, opened 2024-07-31
Evaluate the context file in the database directory when validating

JamesWrigley:
The `dir` argument to NamedTemporaryFile was accidentally deleted in 05cf572, causing context files that rely on executing in the database directory (e.g. to import local files) to fail validation.

I tried to write a test for this but it was trickier than I expected so I opted to just fix it :shrug: Tested on p900460 (CC @daviddoji). This is kinda important because lots of people use local files and otherwise it's not possible to use the editor, so I'll merge this later tonight if there's no objections.


tmichela:
LGTM!

takluyver:
Sorry, my mistake, thanks for the quick fix. :+1:

PR #302, opened 2024-07-31
Fix reprocessing from the GUI when the context dir is not the CWD

takluyver:
My bad :-)

Thanks @tmichela for catching this.


takluyver:
Deployed

PR #306, opened 2024-08-02
FIX: allow plotting variable with Cell's custom summary value

tmichela:
`Cell` is being used in the field, with custom defined summary value as string the GUI thinks there's nothing to plot, although there's still plottable data saved by the variable.


tmichela:
this only partially fixes plotting Cell data, but it unblock users for now, so I'll merge it later today if there's no comments.
I'll open an issue to have a complete fix.

PR #307, opened 2024-08-05
Gui reprocessing: update variable list from file content

tmichela:
After testing #304 it was a bit unpractical to use for mainly 2 reasons:
- people often modify the context file with other editors
- the list would still contain variables present in the table, but not in the file.

So with this PR, we check the context file when we open the reprocess dialog.
It takes a few seconds in practice, but is loading only if we detect that the context file was modified. We could later add that saving the file from the GUI would also update that list.
I additionaly added a filter for the variable list which helps a lot when the context file contains many vars.

reprocess

### DIFF
--- a/damnit/backend/extract_data.py
+++ b/damnit/backend/extract_data.py
@@ -81,8 +81,8 @@ class ContextFileUnpickler(pickle.Unpickler):
         else:
             return super().find_class(module, name)
 
-def get_context_file(ctx_path: Path, context_python=None):
-    ctx_path = ctx_path.absolute()
+def get_context_file(ctx_path: str | Path, context_python=None):
+    ctx_path = Path(ctx_path).absolute()
     db_dir = ctx_path.parent
 
     if context_python is None:
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -345,6 +345,7 @@ da-dev@xfel.eu"""
         self.context_dir_changed.connect(lambda _: self.action_export.setEnabled(True))
         self.action_export.triggered.connect(self.export_table)
         self.action_process = QtWidgets.QAction("Reprocess runs", self)
+        self.action_process.setShortcut("Shift+R")
         self.action_process.triggered.connect(self.process_runs)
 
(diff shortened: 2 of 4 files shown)


tmichela:
I now parse the context file ast instead of executing it, which is much faster (a few ms).

JamesWrigley:
I'm kinda against parsing the context file because at some point I'd like it to be possible to dynamically create variables, which would require runtime side-effects that wouldn't show up in the AST.

tmichela:
can re reconsider that point once the feature exists?  
It would be easy to change that back, and the benefit of not waiting 20 seconds for a list of variables here outweights a future rework, IMHO.

JamesWrigley:
Ok, sure :+1:

takluyver:
The searchable list is very nice, and superqt looks like a good find. :+1: 

Honestly, I'm not keen on the rest of this PR. Yes, we want to deal with changes saved outside the GUI. But doing this specifically for the reprocess dialog means we've got two different ways of reading the context file (AST vs. execution), two different places in the GUI we're keeping a list of variables,  and two different ways of ordering them for display. This all sounds like a recipe for confusing bugs and inconsistencies down the line.

How urgently do we need to make a change? We could obviously put it in as a shortcut and integrate it better later, but I feel like I spend a lot of time on DAMNIT trying to clean up shortcuts, and it's harder to get code review for refactoring compared to new features, so I'm not exactly enthusiastic about this.

Would you be happy if I try to work on making the GUI more broadly aware of the context file being edited elsewhere, with a view to getting something ready by the end of this week?

takluyver:
I agree doing it faster is always nice, but if we decide parsing the AST is good enough, why keep a completely separate pathway to update the variable list by running the context file in a separate process? Using two separate mechanisms means we can't fully exploit either the speed advantage of a statically parsable context file, or the dynamism of an always-executed file.

takluyver:
I'll take a look at my idea in any case. I think #291 was the bit I was wanting to get in before going further on that path, and that's merged now.

takluyver:
Sorry, I guess I got too heated about this. I'll step back and let someone else review it; please disregard my comments.

PR #310, opened 2024-08-08
fix loading data with webViewer

### LINKED ISSUE #308: Plotly widget not showing figure
I am trying to employ `plotly` but the widget which should show the figure gets stuck on a waiting animation.

My context file:
```python
from damnit_ctx import Variable
import plotly.graph_objects as go

@Variable(title="Plotly")
def get_plotly(run):
    # Sample data
    x = [1, 2, 3, 4, 5]
    y = [10, 14, 12, 16, 15]
    
    # Create a scatter plot
    fig = go.Figure(data=go.Scatter(x=x, y=y, mode='markers'))
    
    # Add titles and labels
    fig.update_layout(
        title='Simple Scatter Plot',
        xaxis_title='X-Axis',
        yaxis_title='Y-Axis'
    )
    
    return fig
```

Created column and mouse-hover preview is quite promising:
image

But when I double-click on the cell, a new widget is opened but stucks on waiting animation. The app itself is working, I can close the widget, the waiting circle is spinning but the figure is never generated.
image

I am using Ubuntu, I tried several ways to open the gui: FastX in browser (Chrome, Firefox), `ssh -X` from terminal, as well as FastX app from https://confluence.desy.de/display/IS/FastX3_258905161.html. I tried also different figure settings all with the same result.

Thank you for your help

tmichela:
closes #308

### DIFF
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import shelve
 import sys
 import time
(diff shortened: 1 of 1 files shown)


takluyver:
It's probably fine, but do we understand what sandbox we're disabling here and what the potential consequences are?

tmichela:
the WebViewer uses chromium for rendering the web pages. This disables the sandbox for the chromium renderer.  
I did not find any alternative to fixing that issue.

takluyver:
Following some discussion in chat, this should be fine, but @RobertRosca wanted to have a look into what's happening.

tmichela:
After having spent another fair amount of time (unsuccessfully) trying to fix that issue, and having confirmation from Robert that the risk implied by this change is very low... I'll merge later today if there are no objections.

takluyver:
Thanks, LGTM

tmichela:
thanks!

PR #312, opened 2024-08-12
Fix: xarray plot cbar range limit

tmichela:
add the range limit for the xarray branch, similar to the image branch.

Was deployed for a while on the beta module, will merge end of day if there's no comment.


takluyver:
LGTM

PR #316, opened 2024-08-16
WIP: Refactor GUI plotting code

takluyver:
I wanted to try to disentangle the huge knot of code for creating & updating plots.

This is nearly there, but I still need to

- [ ] Finish up the logic for updating plots when new data arrives

### DIFF
--- a/damnit/gui/plot.py
+++ b/damnit/gui/plot.py
@@ -38,6 +38,7 @@ class Canvas(QtWidgets.QDialog):
         plot_type="default",
         strongly_correlated=True,
         autoscale=False,
+        summary_values=False,
     ):
         super().__init__()
         self.setWindowFlags(
@@ -55,7 +56,10 @@ class Canvas(QtWidgets.QDialog):
 
         layout = QtWidgets.QVBoxLayout(self)
 
+        self.xlabel = xlabel
+        self.ylabel = ylabel
         self.plot_type = plot_type
+        self.summary_values = summary_values
         is_histogram = self.plot_type != "default"
 
         self.figure = Figure(figsize=(8, 5))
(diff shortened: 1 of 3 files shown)


damnit/gui/plot.py
```diff
@@ -22,22 +22,24 @@
 
 log = logging.getLogger(__name__)
 
-class Canvas(QtWidgets.QDialog):
+CORRELATION_MSG = """\
```
JamesWrigley, line 25: There's a trailing space here that gives me a syntax warning:
```suggestion
CORRELATION_MSG = """\
```

JamesWrigley (changes requested):
I found two issues when playing around with the branch:
- Changing the number of histogram bins doesn't seem to remove the initial histogram plot:
  histogram
- The x/y labels when plotting xarray's are getting clipped by default:
  plot-dim-names

  Maybe it's missing a `tight_layout()` call?

takluyver:
Thanks for catching those! I think I've got them both fixed now.

damnit/gui/plot.py
```diff
@@ -584,25 +631,29 @@ def _button_plot_clicked(self, runs_as_series):
             )
             return
 
-        if self.plot_type == "histogram1D":
-            ylabel = xlabel
+        xlabel = self._combo_box_x_axis.currentText()
+        if histogram := self._toggle_probability_density.isChecked():
```
JamesWrigley, line 635: I don't think `histogram` needs to be assigned?

JamesWrigley (changes requested):

takluyver:
Good catch, fixed the zoom for 1D xarray plots.

JamesWrigley (approved):
LGTM!

takluyver:
Thank you! I appreciate this was a lot of code changes to review for no immediate benefit, so I'm especially grateful for the careful review. Hopefully it's easier to work on the plotting code now.

PR #320, opened 2024-08-22
Fix passing through --var options to cluster job

takluyver:
I noticed when triggering some reprocessing from the GUI that it evaluated everything in a cluster job, even though I had selected just one (non-cluster) variable to reprocess. I overlooked passing through the new `--var` arguments when preparing to run a cluster job.


JamesWrigley (approved):
LGTM!

takluyver:
Thanks!

PR #322, opened 2024-08-29
Show when runs are being processed

takluyver:
Put an icon next to the run number while a processing job is working on a run. The tooltip if you hover over that cell shows more details. This will also mean that new runs appear in the table when the backend starts processing them, instead of when the first job finishes.

Screencast from 2024-08-29 12-35-34

The status column would be another option, but we recently discussed hiding that by default.

### DIFF
--- a/damnit/backend/extract_data.py
+++ b/damnit/backend/extract_data.py
@@ -190,7 +190,7 @@ class RunExtractor(Extractor):
             'hostname': socket.gethostname(),
             'username': getuser(),
             'slurm_cluster': self._slurm_cluster(),
-            'slurm_job_id': os.environ.get('SLURM_JOB_ID', ''),
+            'slurm_job_id': self._slurm_job_id(),
         })
 
     @staticmethod
@@ -200,6 +200,18 @@ class RunExtractor(Extractor):
             return None
         partition = os.environ.get('SLURM_JOB_PARTITION', '')
         return 'solaris' if (partition == 'solcpu') else 'maxwell'
+
+    @staticmethod
+    def _slurm_job_id():
+        # Get Slurm job ID in the same format that squeue uses
+        array_job_id = os.environ.get('SLURM_ARRAY_JOB_ID', '')
+        array_task_id = os.environ.get('SLURM_ARRAY_TASK_ID', '')
+        if array_job_id and array_task_id:
+            # In an array job, e.g. '12380337_308'
+            return f"{array_job_id}_{array_task_id}"
+        else:
+            # Not an array job - just use the regular job ID
+            slurm_job_id = os.environ.get('SLURM_JOB_ID', '')
 
     @property
     def out_path(self):
--- a/damnit/backend/extract_data.py
+++ b/damnit/backend/extract_data.py
@@ -16,6 +16,7 @@ import subprocess
 import sys
 from pathlib import Path
 from tempfile import TemporaryDirectory
+from uuid import uuid4
 
 import h5py
 import numpy as np
@@ -203,6 +204,19 @@ class Extractor:
         if proposal is None:
             proposal = self.db.metameta['proposal']
 
+        processing_id = str(uuid4())
+        self.kafka_prd.send(self.db.kafka_topic, msg_dict(
+            MsgKind.processing_started, {
+                'processing_id': processing_id,
+                'proposal': proposal,
+                'run': run,
+                'data': run_data.value,
+                'hostname': socket.gethostname(),
+                'slurm_cluster': os.environ.get('SLURM_CLUSTER_NAME', ''),
+                'slurm_job_id': os.environ.get('SLURM_JOB_ID', ''),
+            }
+        ))
+
         out_path = Path('extracted_data', f'p{proposal}_r{run}.h5')
         out_path.parent.mkdir(parents=True, exist_ok=True)
         if out_path.parent.stat().st_uid == os.getuid():
@@ -231,6 +245,10 @@ class Extractor:
             self.kafka_prd.send(self.db.kafka_topic, update_msg).get(timeout=30)
 
         log.info("Sent Kafka updates to topic %r", self.db.kafka_topic)
+
+        self.kafka_prd.send(self.db.kafka_topic, msg_dict(
+            MsgKind.processing_finished, {'processing_id': processing_id}
+        ))
 
         # Launch a Slurm job if there are any 'cluster' variables to evaluate
         if not cluster:
--- a/damnit/backend/extract_data.py
+++ b/damnit/backend/extract_data.py
@@ -31,45 +31,15 @@ from .extraction_control import ExtractionRequest, ExtractionSubmitter
 log = logging.getLogger(__name__)
 
 
-def run_in_subprocess(args, **kwargs):
+def prepare_env():
+    # Ensure subprocess can import ctxrunner & damnit_ctx
     env = os.environ.copy()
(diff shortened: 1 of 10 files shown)


JamesWrigley:
I confess I don't really like that the implementation depends on the GUI being running throughout the processing time, it'd be quite confusing to see some runs being processed in one window and them not being processed in another (particularly important for long-running jobs like at SPB and MID). I'd prefer do this properly and save the job state in the database.

Other things:
- If I read the code correctly this will show a job still running even if the slurm job was preempted, which is quite possible on `exfel`. We should periodically (every ~2 minutes or so) check slurm to see if the job is still alive.
- This also doesn't account for multiple jobs running simultaneously for the same run, but I think we can overlook that for now since we don't really support it anyway.

JamesWrigley:
We should also have at least some tests :sweat_smile:

takluyver:
I know it doesn't work correctly for GUIs launched while a run is already processing, but I chose to do it this way because:

- I'm trying to avoid writing to the DB when every extractor process starts, because the thundering herd can hit database timeouts waiting for the lock.
- We probably want the Kafka messages anyway, for prompt updates, so I don't think this makes it any harder to do a better implementation later.
  - It might even be possible to use Kafka's own persistence, rewinding a few hours and replaying messages from before the GUI started. Not sure about this yet.
  - Another option would be for the `extract_data` machinery to send out 'still running' messages every couple of minutes while the child process runs the context file, so any newly launched GUIs gets the correct state after a while.
- From what I've seen at FXE, it's normal to leave windows open for long periods, so the simple implementation is already useful.

> this will show a job still running even if the slurm job was preempted, which is quite possible on exfel. We should periodically (every ~2 minutes or so) check slurm to see if the job is still alive.

Yup, makes sense.

JamesWrigley:
The main reason I'm against this implementation is that it can lie to the user. That's a hard blocker for me, we should avoid displaying wrong information at all costs. A secondary reason is that we need the same functionality for the web interface, so I'd prefer if we didn't implement something that will specifically only work for the Qt GUI.

> * From what I've seen at FXE, it's normal to leave windows open for long periods, so the simple implementation is already useful.

I would rather not have the feature at all than display potentially wrong information :shrug: Even apart from that, other users doing analysis sometimes open the GUI on fastx so it doesn't really matter if the session on the control hutch PC's is always running.

takluyver:
Would you be OK with this if any incorrect information was time-limited, so it got the right state within, say, 60 seconds after you launch the GUI? Or would that still be unacceptable?

> we need the same functionality for the web interface, so I'd prefer if we didn't implement something that will specifically only work for the Qt GUI.

I'd still build that around the backend sending out messages similar to these, just that some server piece would have to pass them on to the frontend. I don't see this as something that only works for Qt (besides the parts in the GUI code, of course).

takluyver:
OK, already processing runs should show up within 10 seconds, i.e. they send a 'running' message every 10 seconds.

The GUI checks every 2 minutes, as you suggested, for Slurm jobs that have exited without sending a 'finished' message. I think it should also be feasible to catch most types of failure & cancellation and send the message before the job exits, but this is a good backstop in any case.

damnit/backend/extract_data.py
```diff
@@ -197,51 +168,126 @@ def update_db_vars(self):
...
+        super().__init__()
+        self.proposal = proposal
+        self.run = run
+        self.cluster = cluster
+        self.run_data = run_data
+        self.match = match
+        self.variables = variables
+        self.mock = mock
+        self.uuid = str(uuid4())
+        self.running_msg = msg_dict(MsgKind.processing_running, {
```
JamesWrigley, line 183: Could we send and display the user it's running under as well?

damnit/gui/table.py
```diff
@@ -627,6 +633,96 @@ def handle_variable_set(self, var_info: dict):
...
+                if job_id := info['slurm_job_id']:
+                    msg += f" (Slurm job {job_id})"
+                runnr_item.setToolTip(msg)
+            else:
+                runnr_item.setToolTip(f"Processing in {len(jobs_for_run)} jobs")
+        else:
+            runnr_item.setData(f"{run}", Qt.ItemDataRole.DisplayRole)
+            runnr_item.setToolTip("")
+
+    def check_slurm_jobs(self):
```
JamesWrigley, line 678: I would recommend putting this in the backend because it'll be necessary for the web interface at some point.

JamesWrigley (changes requested):

JamesWrigley:
Just for context, I would still like to refactor this in the future to save the job status' in the database so we can alert users if a job was preempted or failed for some reason.

damnit/backend/extract_data.py
takluyver: Yup, that makes sense.

damnit/gui/table.py
takluyver: Good point, I've refactored this to not be part of the table model class now. Yesterday I was just trying to make this bit work as quickly as possible.

takluyver:
I think this probably needs a bit of a change now that we're using array jobs, to make sure that the different ways of identifying Slurm jobs line up.

takluyver:
I did the change for Slurm array jobs. I looked into using `squeue --json` to get the job statuses, but it doesn't seem to work with the `--jobs` option & selecting >1 job, giving you a mountain of output for all jobs instead. So I went back to the splitting lines of text approach.

damnit/backend/extract_data.py
```diff
@@ -198,51 +170,139 @@ def update_db_vars(self):
...
+    def _slurm_job_id():
+        # Get Slurm job ID in the same format that squeue uses
+        array_job_id = os.environ.get('SLURM_ARRAY_JOB_ID', '')
+        array_task_id = os.environ.get('SLURM_ARRAY_TASK_ID', '')
+        if array_job_id and array_task_id:
+            # In an array job, e.g. '12380337_308'
+            return f"{array_job_id}_{array_task_id}"
+        else:
+            # Not an array job - just use the regular job ID
+            slurm_job_id = os.environ.get('SLURM_JOB_ID', '')
```
tmichela, line 214:
```suggestion
            return slurm_job_id = os.environ.get('SLURM_JOB_ID', '')
```

Is this branch used at all or all jobs are started as array now?

tmichela (changes requested):
2 visual suggestions:
1. There some delay between the re-processing request and the visual hint of processing appearing, which confused me a bit whether my request was working or not. Can we instead, immediately have a "processing requested" hint ?
e.g. "⏱" > "⚙️" > "✅"/"❌"

3. Can we show the hint on the header rather than the "run" column? This way it is always visible.

takluyver:
I've added an icon for queued jobs. Whatever font the UI uses doesn't have symbols for hourglass or stopwatch (& I'm not sure a tiny stopwatch would be distinctive), so I've gone with vertical ellipsis (⋮) for now.

And I've put the symbol in the row header as you suggest. This does mean the width of the headers changes as icons appear & disappear, which isn't super elegant, but probably not a big deal.

Screencast From 2025-01-22 16-43-59

I'd like to leave success/error indicators for later, if that's OK with you. I'd ideally like to indicate errors calculating specific variables in the relevant cells.

tmichela:
> Whatever font the UI uses doesn't have symbols for hourglass or stopwatch

Looks like this is a bold text issue, but the dots are fine too.

grafik

> I'd like to leave success/error indicators for later, if that's OK with you.

Sure.

LGTM!

tmichela (approved):

takluyver:
Thank you!

PR #323, opened 2024-08-30
fallback on local execution if slurm fails

tmichela:
This is mainly to enable external colleagues to start testing DAMNIT without access to slurm.  
It would also be useful later this year if #255 fails dramatically for reasons...

### DIFF
--- a/damnit/backend/listener.py
+++ b/damnit/backend/listener.py
@@ -31,6 +31,26 @@ KAFKA_CONF = {
 
 log = logging.getLogger(__name__)
 
+# tracking number of local threads running in parallel
+# only relevant if slurm isn't available
+MAX_CONCURRENT_THREADS = min(os.cpu_count() // 2, 10)
+local_extraction_threads = []
+
+
+def execute_direct(submitter, request):
+    for th in local_extraction_threads.copy():
+        if not th.is_alive():
+            local_extraction_threads.pop(local_extraction_threads.index(th))
+
+    if len(local_extraction_threads) >= MAX_CONCURRENT_THREADS:
+        log.warning(f'Too many events processing ({MAX_CONCURRENT_THREADS}), '
+                    f'skip event (p{request.proposal}, r{request.run}, {request.run_data.value})')
+        return
+
+    extr = Thread(target=submitter.execute_direct, args=(request, ))
+    local_extraction_threads.append(extr)
+    extr.start()
+
 
 class EventProcessor:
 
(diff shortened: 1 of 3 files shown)


damnit/backend/listener.py
```diff
@@ -110,7 +111,11 @@ def handle_event(self, record, msg: dict, run_data: RunData):
         log.info(f"Added p%d r%d ({run_data.value} data) to database", proposal, run)
 
         req = ExtractionRequest(run, proposal, run_data)
-        self.submitter.submit(req)
+        try:
+            self.submitter.submit(req)
+        except Exception:
+            log.warning("Slurm job submission failed, starting process locally")
```
takluyver, line 117: Might be worth logging some info about the exception here, e.g. `exc_info=True` to include a traceback in the logs. We could suppress that for `FileNotFoundError`, which is what people should see without Slurm available.

tmichela: good point, thanks

JamesWrigley:
Isn't this a bit dangerous? It could lead to the thundering herd problem if submitting slurm jobs fails for some reason. It also won't apply to `cluster=True` variables.

tmichela:
I think the risk is *very* low. This goes back to the previous behavior, which has seen issues only one week (IIRC) when there were kafka server connection issues. And, this fallback should™️  never happen in the regular case, or rather many other things will break if there's an issue with slurm.  
I can track the number of running threads with an arbitrary limit, if you're still too worried about the risk. Do you prefer to queue or drop new runs if this scenario happens?

About the cluster variables, I believe we don't want to have them running locally, at least from the listener process, right? So, I think the current behavior will just have the submitter to fail when trying to start the slurm job. I could add a log message if that's what you mean?

JamesWrigley:
Ok, sure :+1: 

> About the cluster variables, I believe we don't want to have them running locally, at least from the listener process, right? So, I think the current behavior will just have the submitter to fail when trying to start the slurm job. I could add a log message if that's what you mean?

I was mostly thinking of the external colleagues, but if they don't want to test the cluster variables then yeah we can just let them fail. Otherwise we could add a database option to make slurm a noop.

tmichela:
I've added a limitation of local runners anyway, and unit test. I'll merge tomorrow if there are no more comments.

> I was mostly thinking of the external colleagues, but if they don't want to test the cluster variables then yeah we can just let them fail.

Since they want to run it in an environment without access to a cluster with slurm, I assume it's clear cluster variables aren't supported. But we can see and add a more obvious options if that become an issue.

PR #324, opened 2024-08-30
Hide the Status column

tmichela:


takluyver:
I think 'Status' can now also be removed from here:

https://github.com/European-XFEL/DAMNIT/blob/88d2e751f028eb0068f0b9906a3bee10e5de4e16/damnit/gui/table.py#L177-L179

Other than that, LGTM - I assume you've done some manual testing to see that nothing is obviously completely broken in the GUI.

PR #327, opened 2024-09-02
Document how to install the API

JamesWrigley:
CC @CammilleCC


CammilleCC:
I forgot that we have a PyPi entry, thanks for the hint and for the docs update! LGTM

PR #328, opened 2024-09-02
Add support for optionally returning the Plotly JSON string

JamesWrigley:
This is useful for the web interface because the JSON string needs to be sent to the browser.

CC @CammilleCC


JamesWrigley:
(sneakily bumped the API version in 3aaed41 so this'll hit pypi after merging)

CammilleCC:
Thanks for the quick PR, James! LGTM

PR #331, opened 2024-09-11
Add changelog to documentation

tmichela:
Adds a changelog page to docs since we now version and publish the package.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,6 +2,8 @@
 
 ## [0.1.4]
 
+2024-09-03
+
 Only API change: added a deserialize_plotly option to read() to allow reading the Plotly JSON string.
 
 Added:
@@ -24,6 +26,8 @@ Fixed:
 [Full Changelog](https://github.com/European-XFEL/DAMNIT/compare/0.1.3...0.1.4)
 
 ## [0.1.3]
+
+2024-07-25
 
 Added:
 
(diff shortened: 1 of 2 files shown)


docs/changelog.md
```diff
@@ -0,0 +1,174 @@
...
+- Plot xarray (!289).
+
+Changed:
+
+- Preserve context file path for tracebacks (!249).
+- Improve error handling when opening variables in the GUI (!257).
+- Evaluate non-cluster variables in Slurm using Solaris cluster (!256).
+- Use Slurm jobs for reprocessing (!270).
+- HDF5 array compression (!273).
+- Simplify testing CLI by passing argv in (!279).
```
takluyver, line 51: I generally think changelogs should describe user-visible changes, so I tend to leave out things about testing, CI, and most docs changes. YMMV :slightly_smiling_face:

takluyver:
Are we aiming to record application changes going forwards? Or only changes affecting the API, on the grounds that that's why we started doing numbered versions, and we routinely deploy the application between 'releases'?

docs/changelog.md
tmichela: right, I did remove quite a bit already, some may have slip through. I cleaned a bit more now

tmichela:
I would be in favor to recording application changes, as people primarily go to the documentation for the GUI usage, how to write a context file and how to run the backend.

takluyver:
OK, that makes sense. Is it worth also putting the date each version was released, so you can get a rough sense of when changes were made? E.g. I [do this for EXtra-data](https://extra-data.readthedocs.io/en/latest/changelog.html).

Other than that, LGTM

tmichela:
Thanks, added the release dates.

PR #332, opened 2024-09-12
Activate software opengl by default on maxwell nodes.

tmichela:
Since this cause a few support request already, this activates software openGL by default on maxwell when starting the gui on a non-display node.


tmichela:
I'll merge and deploy later today if there's no objection.

takluyver:
LGTM

PR #334, opened 2024-09-13
Add back grid lines for DataArray plots

JamesWrigley:
I think these were lost with the big refactor.


tmichela:
I probably forgot that when I branched the xarray case, LGTM

damnit/gui/plot.py
```diff
@@ -423,6 +423,10 @@ def __init__(self, parent, data, **kwargs):
             self._nan_warning_label.show()
 
         data.plot(ax=self._axis)
+
+        if data.ndim == 1:
```
takluyver, line 427: I don't think you need the ndim check here - this class is specific to 1D dataarrays, and others will go into `ImagePlotWindow`.

JamesWrigley: Good point, removed in b799785.

PR #335, opened 2024-09-13
Use Slurm array jobs to limit concurrent extraction jobs

### LINKED ISSUE #96: Consider using slurm job arrays
When reprocessing lots of runs with e.g. `amore-proto reprocess all`, we always loop over each run one at a time and launch a slurm job for each one. For a proposal with 323 runs this takes about 40 minutes, which is a bit annoying if you only want to reprocess cluster variables anyway.

An alternative would be to launch all the slurm jobs at once at the beginning using job arrays: https://slurm.schedmd.com/job_array.html
And then loop over the runs to process the non-cluster variables as usual.

takluyver:
When reprocessing a whole bunch of runs at once, submit them as a Slurm array job, and ask Slurm to limit how many will run at once. This should mitigate the 'database locked' errors, without requiring people to manually batch their jobs and monitor their progress.

The limit is set at 30 concurrent jobs for now. Obviously we can make that configurable.

I've taken the shortcut of using the array task IDs for the run numbers. This avoids having to define a way to communicate 'array task 1 -> run 53', but it does mean that only the run numbers can vary within an array. So the one job we ask to update the variables table in the DB is submitted separately, and if you do `reprocess all` on a database spanning multiple proposals, each proposal would be submitted separately, raising the effective limit to N * 30 jobs. That's not ideal, but so far it's largely theoretical.

Closes #96.

### DIFF
--- a/damnit/backend/extraction_control.py
+++ b/damnit/backend/extraction_control.py
@@ -196,6 +196,10 @@ class ExtractionSubmitter:
 
             script_expr = f".tmp/launch-{grpid}-$SLURM_ARRAY_TASK_ID.sh"
             cmd = self.sbatch_array_cmd(script_expr, req_group, limit_running)
+            if out:
+                # 1 batch at a time, to simplify limiting concurrent jobs
+                prev_job = out[-1][0]
+                cmd.append(f"--dependency=afterany:{prev_job}")
             res = subprocess.run(
                 cmd, stdout=subprocess.PIPE, text=True, check=True, cwd=self.context_dir,
             )
--- a/damnit/backend/extraction_control.py
+++ b/damnit/backend/extraction_control.py
@@ -10,9 +10,9 @@ import subprocess
 import sys
 from contextlib import contextmanager
 from ctypes import CDLL
-from dataclasses import dataclass, replace
-from itertools import groupby
+from dataclasses import dataclass
 from pathlib import Path
+from secrets import token_hex
 from threading import Thread
 
 from extra_data.read_machinery import find_proposal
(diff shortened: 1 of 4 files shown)


takluyver:
:face_exhaling: no, we can't use array task IDs directly as run numbers, because:

> The minimum index value is 0. the maximum value is one less than the configuration parameter MaxArraySize. 

```
MaxArraySize            = 1001
```

So this will need a rework.

takluyver:
OK, now it creates one script per run in `.tmp` under the AMORE directory. The scripts clean themselves up when they run, so we don't get hundreds of tiny files left around. This allows more diverse jobs to be submitted as one array.

damnit/backend/extraction_control.py
```diff
@@ -274,11 +347,11 @@ def reprocess(runs, proposal=None, match=(), mock=False, watch=False, direct=Fal
...
+        for req in reqs:
             submitter.execute_direct(req)
-        elif watch:
+    elif watch:
+        for req in reqs:
             submitter.execute_in_slurm(req)
-        else:
-            submitter.submit(req)
+    else:
+        submitter.submit_multi(reqs, limit_running=limit_running)
```
JamesWrigley, line 357: Is the old `.submit()` used anywhere else, or can we delete it?

JamesWrigley (changes requested):

damnit/backend/extraction_control.py
takluyver: Good question. It's still used inside `extract_data`, when the Solaris job finishes and launches a Maxwell job. We could change that to use `submit_multi` and always make an array job with 1 task - but it would mean the job names in `squeue` are less useful, because they don't include the run number.

(The thing where Solaris jobs launch Maxwell jobs also means we're not strictly limiting how many extraction jobs can run simultaneously. I don't think there's any easy way around that, but I figure that staggering when jobs start should still mitigate the thundering herd anyway.)

JamesWrigley: Alrighty, let's leave it then.

JamesWrigley (approved):
I'm increasingly worried by this swathe of code that's entirely untested :sweat_smile: Maybe in the future we can add some tests that'll only run on Maxwell. Anyway, LGTM!

takluyver:
Good point, I'll try to figure out some ways to test this stuff better. Thanks for the review!

PR #336, opened 2024-09-16
Register Qt window created in backend test for cleanup

takluyver:
I hope this will avoid the random hard crash during tests I've seen a couple of times recently. It seems to come from garbage collection just after `test_custom_environment` (e.g. [this run](https://github.com/European-XFEL/DAMNIT/actions/runs/10852847320/job/30119808537?pr=335)), so this seems like a good candidate.


tmichela:
LGTM!

JamesWrigley (approved):
LGTM!

PR #337, opened 2024-09-17
FIX: failed to setup new database from the GUI

tmichela:
`TableView.set_columns` fails since `status` was added to the static columns, handle case with zero columns while setting a new db.

@CammilleCC


tmichela:
I'll merge and deploy now since it's urgent

PR #338, opened 2024-09-18
Adds mymdc techniques annotation

tmichela:
This adds support for requesting the techniques information from Mymdc, once @RobertRosca deploys the path in ZWOP.

A run may have multiple techniques associated to it. Here I made the choice to return a single string with all techniques rather than a list, as I think most case will only contain a unique technique (as a mater of fact, the DAQ currently support applying only one technique to a run), and it is easier to handle on the user side. But we can change that if you disagree.

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -97,7 +97,7 @@ class MyMetadataClient:
     @_cache
     def _techniques_info(self, run: int) -> dict[str, Any]:
         run_info = self._run_info(run)
-        response = requests.get(f'{self.server}/api/mymed/runs/{run_info["id"]}',
+        response = requests.get(f'{self.server}/api/mymdc/runs/{run_info["id"]}',
                                 headers=self._headers, timeout=self.timeout)
         response.raise_for_status()
         return response.json()['techniques']
@@ -126,7 +126,7 @@ class MyMetadataClient:
         return ', '.join(t['name'] for t in self._techniques_info(run))
 
     def techniques_identifier(self, run: int) -> str:
-        return ', '.join(t['indentifier'] for t in self._techniques_info(run))
+        return ', '.join(t['identifier'] for t in self._techniques_info(run))
 
 
 class ContextFileErrors(RuntimeError):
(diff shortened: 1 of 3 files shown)


tmichela:
We could also just add a single argument `mymdc#techniques` returning a list containing `name`, `identifier`, and `url` for all techniques. Not sure what's the most useful

takluyver:
How about we offer the singular forms, `technique_name` & `technique_identifier` (or `id`, as a commonly understood shorthand :shrug: ) for now, either taking only the first one, or validating that there's exactly 1 technique. Then if people start recording multiple techniques for a run,  we can add the plural forms, to return lists.

I don't feel strongly about this, but if there's usually 1 technique, it's easy to miss that you're getting a comma-separated list and write something like `if technique == 'WAXS'` when `if 'WAXS' in techniques` is more like what you mean.

tmichela:
> if there's usually 1 technique

That's the key question here, and I don't know the answer to it. And since it's new I think it's not used yet (maybe @dallanto  has an idea?).  
But singular makes sense, unless having more than one technique per run is not unusual.  
Alternatively, we could have a simple `techniques` returning a list of dict (just the information as it is stored on mymdc).

takluyver:
I think it's not unusual for FXE to do two types of measurement in the same run - they definitely have many proposals with two techniques, e.g. XES & XAS, and I think the aim is generally to do both simultaneously.

I found runs in [p900463](https://in.xfel.eu/metadata/proposals/1134#proposal-runs) with techniques recorded, including one 'Test DAQ' run with 2 techniques: [r203](https://in.xfel.eu/metadata/proposals/1134/runs/252177).

So maybe it is worth doing the plural version. In that case I'd go with lists, or a list. I don't have a strong preference between one list-of-dicts vs lists-of-strings. Other than the name & PaNET identifier, I guess you might conceivably want the myMdC ID, and maybe the PaNET URL if you're going to get all semantic web with it. Neither seem especially likely, though I guess myMdC could grow new fields in the future.

tmichela:
I've changed it for now to return the entire technique information as it is stored in mymdc (list of dicts), that's probably the most versatile and clean option. One can easily iterate over elements and get the information required.  
I considered returning an object, that would directly render nicely if returned directly, and would offer easy way of accessing the data, but I think I don't like that idea too much.

dallanto:
Hi Thomas & Thomas, "entire information" like all techniques assigned to a given run, e.g. 2 as in the 900463:r203 case?  
I agree to this: then "techniques" in plural and list of dicts.

takluyver:
LGTM, thanks

tmichela:
Thanks, I'll merge once ZWOP is updated (@RobertRosca 👀 )

RobertRosca:
So the techniques endpoint isn't needed, only the `/runs/{id}` endpoint, right? If so then you can test this with the staging instance at https://exfldadev01.desy.de/zwop-staging/

PR #339, opened 2024-09-18
Lower the default slurm concurrency limit to 15

JamesWrigley:
With 30 I still observe some SQLite locking errors, including from the GUI (which crashed it).


takluyver:
If we're changing the default, let's change it for the CLI as well:

https://github.com/European-XFEL/DAMNIT/blob/2548d5ebe39ee9e89beb6dd5268e9d151162ea1f/damnit/cli.py#L112-L115

Other than that, LGTM.

JamesWrigley:
Roger that, fixed in 5ebdc47.

PR #349, opened 2024-10-24
Set the log file permissions *before* shutting down

JamesWrigley:
The previous logic was not particularly useful...

Unfortunately this is a bit difficult to automatically test so I tested it manually.


takluyver:
LGTM, thanks

PR #354, opened 2024-11-19
Add Variable tagging

tmichela:
Add an option to associate tags to Variables.
The gui code is just a simple filter toggle for now, I plan to modify it once the row filtering feature is finished.

[](https://github.com/user-attachments/assets/76beaa83-f331-4db0-87bd-b74e91d12eaa)


hcraT (commented):
Only very minor staff so I think the PR is ready to merge.

I have just one doubt:

It could totally be that I've missed it but, it seems that there is no automatic update of the selection upon changes in the context file (i.e. a tag that is selected/unselected is added/removed from a variable or a tag that was selected gets removed from the db). The only way to refresh the visibility status is the user manually interacting with the filter (i.e. actually selecting/deselecting some tag). I guess that this could work in most cases but could also be confusing for some user that maybe expect changes in the context file to be immediately reflected in the status of the UI.

damnit/backend/db.py
```diff
@@ -245,6 +258,19 @@ def update_computed_variables(self, vars: dict):
                 for (n, v) in updates.items()
             ])
 
+            for name, var in updates.items():
+                existing_tags = set(self.get_variable_tags(name))
+                new_tags = set(var.get('tags', None) or [])
```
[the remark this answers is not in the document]
tmichela: Tags' value can be `None`, and instantiating a `set` from `None` will raise an error, so I use a empty list in this case.

tmichela:
> there is no automatic update of the selection upon changes in the context file

Good point I overlooked that, thanks! the last commit addresses this case

damnit/backend/db.py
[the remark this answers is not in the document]
hcraT:
```suggestion
                new_tags = set(var.get('tags') or [])
```
Yeah, you are right. This should cover all the cases as `None` is already the default value returned from `get`.

damnit/gui/main_window.py
```diff
@@ -533,6 +533,10 @@ def handle_update(self, message):
             )
         elif msg_kind == MsgKind.variable_set:
             self.table.handle_variable_set(data)
+            # update tag filtering
+            self.table_view.apply_tag_filter(
+                self.table_view._current_tag_filter
+            )
```
hcraT, line 539: I think that it would be nice to inform the user about the update, given that is automatic. A simple message in the status box or also through some change of the filter button (little dot or changing the color).

tmichela: I think I'll go against that one. Tagging a variable will happen 99.9% of the time while developing the context file, particularly not while using the GUI. And I can't think of a particularly useful hint for that change. For the visible filters, we already can see the list of variable change and for hidden tags, that's probably not a useful information.
We can of course revisit that once the feature is used if necessary.

hcraT: That's fair. Probably the only case in which a more explicit message would be useful would be when something is actually selected. I guess that while developing one doesn't really filter the variables. My only "concern" is that one gets confused as the shown variables change without an apparent reason (for example if one uses the interface while someone else changes stuff). But I guess that we can wait and see if this is actually confusing for users.

damnit/ctxsupport/damnit_ctx.py
```diff
@@ -64,6 +65,12 @@ def check(self):
             problems.append(
                 f"data={self._data!r} for variable {self.name} (can be 'raw'/'proc')"
             )
+        if self.tags is not None:
+            if not isinstance(self.tags, Sequence) or not all(isinstance(tag, str) for tag in self.tags):
+                problems.append(
+                    f"tags={self.tags!r} for variable {self.name} (must be a string or an iterable of strings)"
```
hcraT, line 71:
```suggestion
            if not isinstance(self.tags, Sequence) or not all(isinstance(tag, str) and tag != "" for tag in self.tags):
                problems.append(
                    f"tags={self.tags!r} for variable {self.name} (must be a non-empty string or an iterable of strings)"
```

I thought of this after doing the review. I think it would be a bit confusing having an empty tag around.

If you agree with this I would also add a test to check for this case.

hcraT:
Great, LGTM!

tmichela:
Thanks a lot for the review @hcraT !
I'll merge tomorrow if there aren't more comments.

PR #355, opened 2024-11-24
Use xarray's plotting methods to generate image thumbnails

JamesWrigley:
This makes the thumbnails match what's displayed when the variable is plotted, particularly when there are coordinates being used as ticks along the axes.

Before (thumbnail is upside down):
image

After:
image


JamesWrigley:
(if there's no objections I'll merge this at the end of the week)

takluyver:
LGTM

PR #356, opened 2024-11-25
Handle get_context_file() throwing an exception in the editor

JamesWrigley:
Previously this would throw an exception if running the `ctxrunner` module failed for some reason, which would cause the whole GUI to crash. Now we show the traceback and a suggestion to contact DA/the DOC.

Note: I added the `ContextFileCheckerThread.context_getter` hack because mocking didn't seem to work in the new thread.

CC @aleonau


JamesWrigley:
I'm not sure why codecov says that block of code isn't covered, if I insert a print statement locally I see that it is executed. And the test wouldn't pass if it didn't execute.

JamesWrigley:
(if there's no objections I'll merge this at the end of the week)

damnit/gui/editor.py
```diff
@@ -51,7 +57,16 @@ def run(self):
...
-                ctx, error_info = get_context_file(ctx_path, self.context_python)
+                try:
+                    ctx, error_info = self.context_getter(ctx_path, self.context_python)
+                except:
+                    # Not a failure to evalute the context file, but a failure
+                    # to *attempt* to evaluate the context file (e.g. because of
+                    # a missing dependency).
+                    help_msg = "# This is a partial error, please check the terminal for the full error message and ask for help from DA or the DOC."
+                    traceback_str = f"{help_msg}\n\n{traceback.format_exc()}"
+                    self.check_result.emit(ContextTestResult.ERROR, traceback_str, 0, 0, "")
```
takluyver, line 68:
```suggestion
                    self.check_result.emit(ContextTestResult.ERROR, traceback_str, 0, 0, self.code)
```

The string of code checked is used as part of saving, so the downstream logic knows that the check has failed on the code we were planning to save. Doesn't do much at the moment, but it's easy to preserve.

JamesWrigley: Fixed in 2cd5163.

takluyver:
Not sure why mocking wouldn't work with QThread. :confused: I can take a closer look if you like, but it's also OK to live with that.

It seems like a known issue that coverage doesn't capture code run by a QThread: https://github.com/nedbat/coveragepy/issues/686 . There are some workarounds there - amounting to calling `sys.settrace(threading._trace_hook)` inside the thread when it starts - if you like, but I'm happy for coverage to be a loose guide, not a hard rule.

JamesWrigley:
Ah, interesting :thinking: I added that line to `run()`, if that's all coverage needs then we might as well use it.

JamesWrigley:
Ok, that didn't work but maybe I have to call it before the function is called like the examples in that issue. Annnnnyway I'm too lazy to get it working so I'll just smash that merge button :sloth:

takluyver:
Yes, reading that issue more closely, it does sound like tracing won't catch the function you're currently in when calling `sys.settrace()`. Go for it, though.

PR #362, opened 2024-12-11
Basic row filtering

tmichela:
Implements basic row filtering based on column values, there are more requirements for this feature, bu that's already a big enough diff and does already something useful.

One regression: support for standalone comments is dropped in the main table. I implemented another table to view them for the proposal that have used them.

[damnit-row-filtering.webm](https://github.com/user-attachments/assets/9db8f950-c58f-4b51-b261-fb40abf0f539)

### DIFF
--- a/damnit/gui/table.py
+++ b/damnit/gui/table.py
@@ -18,7 +18,7 @@ log = logging.getLogger(__name__)
 
 ROW_HEIGHT = 30
 THUMBNAIL_SIZE = 35
-COMMENT_ID_ROLE = Qt.ItemDataRole.UserRole + 1
+
 
 class TableView(QtWidgets.QTableView):
     settings_changed = QtCore.pyqtSignal()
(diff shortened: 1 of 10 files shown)


damnit/gui/table.py
```diff
@@ -359,7 +368,30 @@ def apply_tag_filter(self, tag_names: set):
...
+        return [self._tag_filter_button, self._filter_status]
+
+    def show_horizontal_header_menu(self, position):
+        pos = QCursor.pos()
+        index = self.horizontalHeader().logicalIndexAt(position)
+        menu = QMenu(self)
+        sort_asc_action = QAction(icon(FA6S.arrow_up_short_wide), "Sort Ascending", self)
+        sort_desc_action = QAction(icon(FA6S.arrow_down_wide_short), "Sort Descending", self)
+        filter_action = QAction(icon(FA6S.filter), "Filter", self)
+        hide_action = QAction(icon(FA6S.eye_slash), "Hide column", self)
```
JamesWrigley, line 380: I'd remove the `hide_action` code if it's not going to be finished in this PR.

JamesWrigley (changes requested):
This is more code than I can properly review in one go, but I tested it locally and it works really nicely :tada: One odd thing I noticed is that the comment column can't be filtered categorically? That seems a bit odd because other string columns can be and the comment column is basically a string column.

Also, please either do a squash merge or clean up the commits before merging. I semi-regularly do bisects and git blames and these kind of commits really hinder that :sweat_smile:

tmichela:
> please either do a squash merge or clean up the commits before merging

I cleaned up the commit history

> I noticed is that the comment column can't be filtered categorically?

Oh, well spotted, I get all `UserRole` data, but the data on the comment column is `EditRole`, I'll add a fix for that.

JamesWrigley (approved):
LGTM!

tmichela:
Thanks!

PR #370, opened 2025-01-20
FIX: handle failure generating summary data

tmichela:
Failure to generate a summary data would fail writing the entire HDF5 file for that run.


takluyver:
LGTM

PR #374, opened 2025-01-22
feat: add support for complex number

tmichela:
HED needs to save complex number for their next experiment.
h5py does support saving complex numbers, but netCDF does not yet, there's an option in xarrays's `to_netcdf` to save "invalid netCDF" files allowing these features, c.f. https://docs.xarray.dev/en/stable/user-guide/io.html#invalid-netcdf-files

I understand [this](https://github.com/HDFGroup/hdf5/pull/4630) PR in HDF5 would allow support for these data types in netCDF, but it's not clear when it will be available.


damnit/ctxsupport/ctxrunner.py
```diff
@@ -621,12 +621,18 @@ def save_hdf5(self, hdf5_path, reduced_only=False):
...
+            # about netcdf compatibility since we offer an API to access the
+            # data, so we use the `invalid_netcdf` option to be able to use these
+            # features.
             obj.to_netcdf(
                 hdf5_path,
                 mode="a",
                 format="NETCDF4",
                 group=name,
                 engine="h5netcdf",
+                invalid_netcdf=True,
```
takluyver, line 635: The h5netcdf docs [only list booleans & references](https://h5netcdf.org/#invalid-netcdf-files) as invalid types, and in my brief test it seemed happy to save complex numbers without this flag.

In the absence of a native compound type in HDF5, h5py stores it as a compound type with fields `r` & `i`, and recognises this when reading data back. That may well mean that other NetCDF libraries read it as a compound (struct) type.

tmichela: what's your environment? In the current `amore` environment I get a:

```python
CompatibilityError: complex dtypes are not a supported NetCDF feature, and are not allowed by h5netcdf unless invalid_netcdf=True.
```

damnit/backend/db.py
```diff
@@ -79,10 +80,27 @@ def identify(cls, blob: bytes):
...
+            return cls.complex
 
         return cls.unknown
 
 
+def complex2blob(data: complex) -> bytes:
+    # convert complex to bytes
+    real = data.real.hex()
+    imag = data.imag.hex()
+    return f"_DAMNIT_COMPLEX_{real}_{imag}".encode()
```
takluyver, line 93: If we're storing complex numbers as a binary blob, we could store the actual binary data rather than converting to & from a hex representation:

```python
np.complex128(data).tobytes()
np.frombuffer(b, dtype=np.complex128)

# or

struct.pack("<dd", data.real, data.imag)
struct.unpack("<dd", b)
```

takluyver:
Other than those things, this LGTM

damnit/backend/db.py
tmichela: changed in bcc38c6d4b6cc99eee50803234ad676a3158061d

damnit/ctxsupport/ctxrunner.py
tmichela: thanks, I updated the minimum requirement for h5netcdf

tmichela:
I just noticed that the `run_variables` table has a `summary_type` column that isn't used yet, but sounds like its purpose would be exactly handling this kind of object(?)
@takluyver  you think it'd be better to use that column to define the summary type instead of adding a header to the blob data?

tmichela:
I made the change anyway, let me know what you think

takluyver:
Thanks, I think that makes sense. I had envisaged the summary_type for where there was a more specialised type based on an SQLite type that we also use, e.g. timestamps stored as floats, but I don't see any problems with using it for custom bytes types as well.

LGTM

CammilleCC:
+1 to putting the complex data type to `summary_type` 🥺

tmichela:
thanks both for the review!

PR #376, opened 2025-01-27
Feat: dark theme

tmichela:
Screenshot from 2025-01-27 11-03-52
Screenshot from 2025-01-27 11-04-54

### DIFF
--- a/damnit/gui/editor.py
+++ b/damnit/gui/editor.py
@@ -162,10 +162,6 @@ class Editor(QsciScintilla):
         self._lexer.setColor(colors['string'], QsciLexerPython.SingleQuotedString)
         self._lexer.setColor(colors['number'], QsciLexerPython.Number)
 
-        # Set paper (background) for all styles
-        for style in range(128):  # QScintilla uses style numbers 0-127
-            self._lexer.setPaper(colors['background'], style)
-
         # Apply the new lexer
         self.setLexer(None)  # Clear the old lexer
         self.setLexer(self._lexer)  # Set the new lexer
--- a/damnit/gui/editor.py
+++ b/damnit/gui/editor.py
@@ -15,7 +15,7 @@ from pyflakes.api import check as pyflakes_check
 from ..backend.extract_data import get_context_file
 from ..ctxsupport.ctxrunner import extract_error_info
 from ..context import ContextFile
-
+from .theme import Theme
 
 class ContextTestResult(Enum):
     OK = 0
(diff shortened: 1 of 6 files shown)


damnit/gui/main_window.py
```diff
@@ -1030,6 +1052,55 @@ def show_adeqt(self):
             self.adeqt_window = AdeqtWindow(ns, parent=self)
         self.adeqt_window.show()
 
+    def _toggle_theme(self):
+        """Toggle between light and dark themes."""
+        new_theme = Theme.DARK if self.dark_mode_action.isChecked() else Theme.LIGHT
```
takluyver, line 1057: The triggered signal includes a checked parameter if you want it: https://doc.qt.io/qt-5/qaction.html#triggered

```suggestion
    def _toggle_theme(self, checked):
        """Toggle between light and dark themes."""
        new_theme = Theme.DARK if checked else Theme.LIGHT
```

damnit/gui/editor.py
```diff
@@ -123,6 +121,94 @@ def __init__(self):
...
+            self._lexer.setDefaultPaper(QColor('white'))
+            self._lexer.setDefaultColor(QColor('black'))
+            self._lexer.setColor(QColor('#0000ff'), QsciLexerPython.Keyword)
+            self._lexer.setColor(QColor('#007f7f'), QsciLexerPython.ClassName)
+            self._lexer.setColor(QColor('#7f0000'), QsciLexerPython.Operator)
+            self._lexer.setColor(QColor('#007f00'), QsciLexerPython.FunctionMethodName)
+            self._lexer.setColor(QColor('#7f7f7f'), QsciLexerPython.Comment)
+            self._lexer.setColor(QColor('#7f007f'), QsciLexerPython.DoubleQuotedString)
+            self._lexer.setColor(QColor('#7f007f'), QsciLexerPython.SingleQuotedString)
+            self._lexer.setColor(QColor('#007f7f'), QsciLexerPython.Number)
```
takluyver, line 191: Do we want to have all the colours defined in theme.py, and then have this method do something like:

```python
colours = get_highlighting_colours(self.current_theme)

...
self._lexer.setColor(colours['number'], QsciLexerPython.Number)
```
```

Up to you - I don't feel strongly about how we define the colours, it just seemed worth asking the question.

takluyver:
I don't know much about Qt's styling APIs, but this LGTM.

damnit/gui/editor.py
tmichela: Good point, I moved the color definition to `theme.py`

tmichela:
Thanks!

PR #377, opened 2025-01-27
fix: interpret complex numbers from the DB when the gui loads the table

tmichela:


takluyver:
LGTM

tmichela:
thanks, I also added the missing changes to the changelog

PR #378, opened 2025-01-30
Dark theme: add all lexers and the matplotlib toolbar

tmichela:


tmichela:
I'll merge later today, unless someone shoots.

PR #379, opened 2025-01-30
FEAT: transient variables

tmichela:
The controversial one 🙈 

Transient variables' data and summary are not saved.

### DIFF
--- a/docs/backend.md
+++ b/docs/backend.md
@@ -71,9 +71,9 @@ these arguments:
   ```
 - `cluster` (bool): whether or not to execute this variable in a Slurm job. This
   should always be used if the variable does any heavy processing.
-- `transient` (bool): whether or not to save the variable's result to the
-  database. This is useful for e.g. intermediate results. By default variables
-  save their results (transient=False).
+- `transient` (bool): do not save the variable's result to the database. This
+  is useful for e.g. intermediate results to be reused by other Variables. By
+  default Variables do save their results (transient=False).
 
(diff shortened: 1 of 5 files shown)


docs/backend.md
```diff
@@ -71,6 +71,9 @@ these arguments:
   ```
 - `cluster` (bool): whether or not to execute this variable in a Slurm job. This
   should always be used if the variable does any heavy processing.
+- `transient` (bool): whether or not to save the variable's result to the
+  database. This is useful for e.g. intermediate results. By default variables
```
takluyver, line 75: "whether or not to save" implies that `True` means save.

I imagine most users will be familiar with the word transient anyway, so the flip will be obvious, but maybe we could either reword this ("whether or not to skip saving"), or change the sense of the parameter (`saved=False`).

takluyver:
Other than quibbling over naming & documentation, this LGTM

docs/backend.md
tmichela: do you have a preference? `transient` vs `saved`

takluyver: Maybe a slight meaning towards `saved`, which is maybe clearer about the difference from regular variables. But I assume most variables will be saved, so calling the exceptions 'transient' rolls off the tongue better than 'unsaved'. :shrug:

tmichela: OK, I clarified the docs, and kept `transient`.

PR #386, opened 2025-02-07
feat: Rename CLI from 'amore-proto' to 'damnit'

tmichela:


takluyver:
LGTM, thanks!

(I like that the `listen` subcommand is now "Damnit, listen!")

PR #390, opened 2025-02-11
Don't pass DISPLAY environment variable to Slurm jobs

takluyver:
Should fix issues with interactive plots from jobs popping up in FastX, and hopefully also issues with jobs from the listener hanging.


JamesWrigley (approved):
I also haven't tested it, buuut I don't see any downside so LGTM!

takluyver:
Thanks!

PR #392, opened 2025-02-12
feat: column filter indicator

tmichela:
Adds a visual indicator on columns with active filter:
grafik


JamesWrigley (approved):
LGTM! (but it's missing a changelog entry)

tmichela:
Thanks! Changelog entry is added.

PR #399, opened 2025-02-18
Handle run timestamps from old proposals with only proc files used

JamesWrigley:
I've sneakily already hotfixed this in the beta module :shushing_face: It would be more correct to find the raw files buuut I'm lazy and don't care much about old proposals anyway.

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -374,6 +374,12 @@ def get_start_time(xd_run):
         # If the timestamp information is not present (i.e. on old files), then
         # we take the timestamp from the oldest raw file as an approximation.
         files = sorted([f.filename for f in xd_run.files if "raw" in f.filename])
+        if len(files) == 0:
+            # Some old proposals also copy all the non-detector data into proc,
+            # in which case the DataCollection will only open the proc files. In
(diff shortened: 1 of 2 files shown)


damnit/ctxsupport/ctxrunner.py
```diff
@@ -374,6 +374,12 @@ def get_start_time(xd_run):
         # If the timestamp information is not present (i.e. on old files), then
         # we take the timestamp from the oldest raw file as an approximation.
         files = sorted([f.filename for f in xd_run.files if "raw" in f.filename])
+        if len(files) == 0:
+            # Some old proposals also copy all the non-detector data into proc,
+            # in which case the DataCollection will only open the proc files. In
+            # this case we fall back to the timestamp of the proc files, which
+            # should be pretty close to the timestamp of the raw files.
```
tmichela, line 381: This statement is not necessary true. It's not unusual to have data migrated or (re-)corrected much after it's been collected. I would say especially for older proposals...
Maybe the correct way is to query MyMdc?

But yea 🤷 , I guess it's not too critical, so LGTM.

JamesWrigley: True, mymdc would be the better option.

PR #400, opened 2025-02-18
Feat: numerical filter range selection

tmichela:
Adds a line plot of the data distribution and slider to select filter ranges.

[numericalFilterRange.webm](https://github.com/user-attachments/assets/6f1d41a8-f147-4ce7-91f2-96b6cc97e9a3)


JamesWrigley:
It's a bit odd that when nothing is selected the plot still shows that everything is selected:
image

Would it be difficult to fix that? No worries if not, just a minor thing.

tmichela:
ah, yea the value range and item list are kinda separated (though the list gets updated when the range changes)...
Right now the plot shows the distribution of all values for that column,  I plan for improvement on that, including effect of selections from other filters, but I'd prefer to ignore that for now if that's OK.

damnit/gui/widgets.py
```diff
@@ -260,3 +263,170 @@ def currentLineColor(self, countDistance, totalNrOfLines, trailFadePerc, minOpac
...
+        self.x_data = x_data
+        self.y_data = y_data
+        self.slider_position = (x_data[0], x_data[-1])  # (left, right)
+        self.setMinimumSize(100, 100)
+
+    def set_slider_position(self, position: tuple[float, float]):
+        self.slider_position = position
+        self.update()
+
+    def paintEvent(self, event):
```
takluyver, line 280: I see calls to `.repaint()` in the test code, but it looks like this method isn't getting executed in the tests. Is there a good way to remedy that?

tmichela:
I'll merge tomorrow if there are not additional comments

JamesWrigley (approved):
LGTM!

PR #401, opened 2025-02-20
Fix thumbnail creation for `DataArray`'s with shape `(1, n)`

JamesWrigley:
I opted to stick with a thumbnail for consistency, the idea being that if the users really want a 1D array they should squeeze it themselves.


takluyver:
Thanks, that makes sense to me.

LGTM

PR #402, opened 2025-02-20
Dialog for saving conflicting changes from context file editor

takluyver:
Since #304, the context file editor periodically checks the file for changes and reloads the context file as long as you don't have any unsaved changes locally. This PR checks the file on disk when you try to save, and if it has changed, it tells you this and gives you options to deal with it:

image

Clicking 'View changes' opens a diff viewer (meld) to compare the file on disk (left) with your editor (right). You can pull changes across either way and save the files from there. When you close this view, any changes you saved on the right go back into the editor view.

image

I also wrapped the editor `setText()` method to restore the approximate scroll position when we reload the editor contents. This isn't super smart, but for small edits it works quite nicely, and I'm hoping it's strictly better than jumping up to the top.

### DIFF
--- a/damnit/gui/editor.py
+++ b/damnit/gui/editor.py
@@ -1,3 +1,4 @@
+import logging
 import sys
 import traceback
 from enum import Enum
@@ -8,14 +9,18 @@ from tempfile import NamedTemporaryFile
 from pyflakes.api import check as pyflakes_check
 from pyflakes.reporter import Reporter
 from PyQt5.Qsci import QsciCommand, QsciLexerPython, QsciScintilla
-from PyQt5.QtCore import Qt, QThread, pyqtSignal
+from PyQt5.QtCore import Qt, QProcess, QThread, pyqtSignal
 from PyQt5.QtGui import QFont
+from PyQt5 import QtWidgets
 from superqt.utils import signals_blocked
 
 from ..backend.extract_data import get_context_file
 from ..context import ContextFile
 from ..ctxsupport.ctxrunner import extract_error_info
 from .theme import Theme, ThemeManager, set_lexer_theme
+
+
+log = logging.getLogger(__name__)
 
 
 class ContextTestResult(Enum):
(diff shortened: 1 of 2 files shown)


matheuscteo:
Thanks Thomas, it looks good and worked fine from my side

my only observations are:
 
- Are we sure that `meld` won't crash and be installed? 
- If the GUI crashes or someone closes it from the command line the temp file for the context.py won't be deleted.
- Should we perhaps hide the temp file? I know it exists only for a short time, but wouldn't be nice if a third person tries to edit this one. I'd go for something like it happens for vim does (`.test.swp`)

Apart from that, LGTM

takluyver:
> Are we sure that meld won't crash and be installed?

It's definitely installed on Maxwell, and I think it's probably the most likely graphical diff viewer to be installed on a random Linux system. We could make it configurable, or have a list of different ones to try, but I'm inclined to keep things simple until it becomes an issue.

Meld has been pretty reliable from what I remember of it, but it's possible it could crash. Any error output would be visible in the terminal where you launched DAMNIT.

> Should we perhaps hide the temp file? 

Yup, good idea. :slightly_smiling_face: 

> If the GUI crashes or someone closes it from the command line the temp file for the context.py won't be deleted.

That's true. I don't think there's really a good way to avoid that, but if we let Python create the tempfile in the default `/tmp` folder, then any leftover files will at least be cleaned up when the machine reboots (and potentially sooner, if IT run some temp file cleaner daemon). I don't think these ones actually need to be in the amore folder.

damnit/gui/editor.py
```diff
@@ -189,3 +204,72 @@ def on_test_result(self, res, info, lineno, offset, checked_code):
...
+        cancel = QtWidgets.QPushButton("Cancel", self)
+        cancel.clicked.connect(self.reject)
+        for btn in [self.view_changes, reload, overwrite, cancel]:
+            layout.addWidget(btn)
+        self.view_changes.setFocus()
+
+        self.action = None
+        self.temp_file = NamedTemporaryFile(
+            mode='w+', encoding='utf-8', delete=False,
+            dir=main_window.context_dir, prefix='context-editing-', suffix='.py'
```
takluyver, line 238:
```suggestion
            prefix='.damnit-context-editing-', suffix='.py'
```

PR #403, opened 2025-02-24
Fix filtering of non-cluster variables

JamesWrigley:
Previously these would be included even when `cluster=True`, causing them to be executed twice (once in the non-cluster job and again in the cluster job).

I changed the default value of the `cluster` argument in `Context.filter` to make the tests pass, that should be a safe change because every other caller explicitly passes `cluster`.

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -251,7 +251,7 @@ class ContextFile:
             if not v.transient or inc_transient
         }
 
-    def filter(self, run_data=RunData.ALL, cluster=True, name_matches=(), variables=()):
+    def filter(self, run_data=RunData.ALL, cluster=None, name_matches=(), variables=()):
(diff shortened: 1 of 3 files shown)


takluyver:
I think that was kind of deliberate, but I'm not sure why. Maybe some early version of the filtering didn't keep the dependencies, so the cluster job just re-ran all the non-cluster variables in case they were required. I don't think we need to worry about it, though.

I'd suggest that we have an option - maybe `cluster=None` - to include both `cluster=False` and `=True` variables, since all the other parameters to `.filter()` allow some kind of 'not specified' option. For symmetry, I'd also make that the default, so that e.g. `ctx.filter(run_data=RunData.RAW)` is only doing what's obvious, not also filtering based on cluster-ness.

JamesWrigley:
Roger that, fixed in 5025104.

takluyver:
Thanks, LGTM

PR #404, opened 2025-02-25
Hack/thumbnail data validation

tmichela:
Users have variables producing 3d array (frame, y, x), but want to still display a thumbnail in the table.
To do that, they generate the thumbnail directly in the context file and return a `Cell` object with the `summary_value` being the generated `PNGData` object.
Problem is that testing `isintance(obj, PNGData)` fails in this case, here's a minimal example of what is done:

```python
from damnit_ctx import Variable, Cell

def custom_thumbnail(data):
    from ctxrunner import generate_thumbnail
    image = data.squeeze() if data.squeeze().ndim == 2 else data.mean(dim='trainId')
    png = generate_thumbnail(image)
    cell = Cell(data=data)
    cell.summary_value = png
    return cell

@Variable(title='Images')
def images(run):
    """Standard xarray return, just to double check it works to get data
    """
    kd = run.alias['OptPeter']
    shot = kd.drop_empty_trains()[0]
    return shot.xarray()

@Variable(title='Custom')
def custom(run, ims:'var#images'):
    """3D images with custom thumbnail
    """
    return custom_thumbnail(ims)
```

I'll patch this because it's quite urgent.


JamesWrigley (approved):
I'm inclined to replace this in the future with supporting generating thumbnails from 2D array summaries, but for now LGTM!

tmichela:
Thanks! 

> I'm inclined to replace this in the future with supporting generating thumbnails from 2D array summaries

yes, absolutely

PR #405, opened 2025-02-25
Fix Dockerfile that was missing LICENSE file

takluyver:
The `build` stage of the Dockerfile was trying to create just enough of the source folder to install the dependencies. This broke on a new version of Flit because the LICENSE file wasn't included. But I don't think there's any problem with just copying the whole git repo folder in, like the app stage does later. So this simplfiies it.

@RobertRosca what do we gain by installing all the dependencies in the build stage and then installing DAMNIT itself separately in the app stage? Could we just install it in the build stage and then copy it across as part of `/app`?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,18 +18,10 @@ RUN set -eux; \
     ; \
     apt list --installed $(apt-mark showmanual) > /.apt-deps-build
 
-COPY pyproject.toml /src/pyproject.toml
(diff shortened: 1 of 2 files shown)


tmichela:
Thanks! LGTM
Maybe you can also revert 21baa24643e70e0bdda7df63b43c49fbae9eb1cc where I lower the flit version.

tmichela (approved):

PR #416, opened 2025-03-12
Show errors from executing variables as tooltips on table cells

takluyver:
Revised presentation:

image

The colours distinguish general errors (orange) from `raise Skip("some reason")` and extra_data SourceNameError (light grey). The latter comes up a fair bit when sources aren't saved, so I want coding errors to stand out a bit more.


tmichela:
Have you checked that this does not affect the filter behavior?
Particularly the columns with thumbnails?

PR #418, opened 2025-03-15
Use anti-aliased interpolation for displaying images

JamesWrigley:
This quite dramatically improves the image quality, particularly on images with small features and noise.

Before and after:
image
image

And another example:
image
image

It's such a big improvement I'm gonna sneakily go ahead and merge it :eyes:


JamesWrigley:
Thanks @keeble for the idea :)

PR #419, opened 2025-03-16
Fix concurrent_jobs usage when reprocessing from the GUI

JamesWrigley:
The GUI calls `submit_multi()` directly so we need to check the database in there instead of `reprocess()`.

@turkot, I'm not sure how this worked for you already if you were reprocessing from the GUI :thinking:

### DIFF
--- a/damnit/backend/extraction_control.py
+++ b/damnit/backend/extraction_control.py
@@ -196,10 +196,13 @@ class ExtractionSubmitter:
             '--wrap', shlex.join(req.python_cmd())
         ]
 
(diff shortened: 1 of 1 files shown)


tmichela (approved):
LGTM

turkot:
> The GUI calls `submit_multi()` directly so we need to check the database in there instead of `reprocess()`.
> 
> @turkot, I'm not sure how this worked for you already if you were reprocessing from the GUI 🤔

Hi James, I have been reprocessing only with the cli, so it looks like I've avoided the issue so far :innocent:

PR #425, opened 2025-03-28
Let the API handle old versions of the HDF5 file without .errors

JamesWrigley:
Previously it would unconditionally try to remove the key, which would crash on older files written without it.

This is quite a serious bug so I'm gonna merge and deploy it.


takluyver:
Oops, my bad. Thanks for catching that. :+1:

PR #426, opened 2025-03-31
Don't treat PENDING Slurm jobs as crashed on job monitor check

takluyver:
I noticed that when the cluster is busy and jobs have to wait a while, the 'queued' status indicator would disappear. This was because the check for jobs ending without notification was only considering 'RUNNING' state, not 'PENDING'. This should fix that.


takluyver:
Thanks!

PR #430, opened 2025-04-02
Listener: catch SQLite read-only error, reopen database and retry

takluyver:
I think this should resolve the errors we've seen with the database connection in the listener becoming read-only. If there's an error on the second attempt, the listener will crash & be restarted by supervisor.

I thought about just crashing the listener on any error & letting it be restarted, but that took me down a rabbit hole of managing Kafka state, so I decided to do this the way I'd already thought of for now, and maybe come back to that idea later.

### DIFF
--- a/damnit/backend/listener.py
+++ b/damnit/backend/listener.py
@@ -3,6 +3,8 @@ import json
 import logging
 import os
 import platform
+import sqlite3
+import time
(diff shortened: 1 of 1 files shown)


tmichela:
Thanks, LGTM!

PR #431, opened 2025-04-02
Store & show preview data as well as summary value

takluyver:
This adds another layer of partially reduced data between `data` and `summary`. Summary values are shown directly within the table cells, whereas preview data is shown when you double click. This is meant to get rid of the common pattern of paired columns, with one for data and one for a diagnostic plot. Preview can be given a 1D or 2D array, or a matplotlib/plotly figure. For simple cases, arrays are nice because the plot in the GUI can be zoomed, whereas mpl figures are saved as a static image.

Previews can be provided via `Cell`:

```python
# Option 1: within the variable function
@Variable()
def spectra(run):
    ...
    return Cell(spectra_arr, preview=spectra_arr.mean(dim=('trainId', 'pulseId'))
```

The preview function currently only gets the main result from the variable function. We could give it the run, and we could allow it to depend on other variables, but I kept it simple for now.

If a preview is provided and a summary (value or reduction function) isn't, a thumbnail of the preview is used for the summary. While working on this, I added thumbnails for 1D arrays, generating a small line plot. The thumbnail popup is not the prettiest, but I think it's useful to see these in the table, especially as rendering a line plot to an RGB image and shrinking it down to a thumbnail is almost impossible to see.

E.g. (first column & tooltip from 1D arrays, second column from matplotlib figures):

Screenshot From 2025-04-02 16-13-35

And here's the same data in a plot window from double clicking:

Screenshot From 2025-04-02 16-13-50

### DIFF
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -240,7 +240,7 @@ class VariableData:
         This is intended for use in Jupyter notebooks
         """
         if (obj := self.preview_data()) is None:
-            return
+            raise ValueError("No preview available")
 
         if isinstance_no_import(obj, 'plotly.graph_objs', 'Figure'):
             obj.show()
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -172,12 +172,16 @@ class VariableData:
                 return blob2complex(value)
             return value
 
-    def preview(self, deserialize_plotly=True):
+    def preview_data(self, *, data_fallback=True, deserialize_plotly=True):
         """Get the preview data for the variable
 
         May return a 1D or 2D data array, a 3D RGB(A) arrray, a plotly figure
         object, a str of plotly JSON (with deserialize_plotly=False) or None
         if no preview is available.
+
+        If no preview was specified in the context file, but the returned data
+        meets the conditions for a preview, this can be returned instead.
+        Pass `data_fallback=False` to prevent this.
         """
         with h5py.File(self._h5_path) as f:
             xarray_group = dset = None
@@ -188,14 +192,18 @@ class VariableData:
                     xarray_group = obj.name
                 else:
                     dset = obj
-            else:
+            elif data_fallback:
                 # Implicit: use data as preview if suitable
                 grp = f[self.name]
                 type_hint = self._type_hint(grp)
-                if self._type_hint(grp) is DataType.DataArray:
+                if type_hint is DataType.DataArray:
                     xarray_group = self.name
+                elif type_hint is DataType.Dataset:
+                    return None
                 else:
                     dset = grp['data']
+            else:
+                return None
 
             if xarray_group is not None:
                 for obj in f[xarray_group].values():
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -171,11 +171,12 @@ class VariableData:
                 return blob2complex(value)
             return value
 
-    def preview(self):
+    def preview(self, deserialize_plotly=True):
         """Get the preview data for the variable
 
-        May return a 1D or 2D data array, a 3D RGB(A) arrray, a plotly figure object,
-        or None if no preview is available.
+        May return a 1D or 2D data array, a 3D RGB(A) arrray, a plotly figure
+        object, a str of plotly JSON (with deserialize_plotly=False) or None
+        if no preview is available.
         """
         with h5py.File(self._h5_path) as f:
             xarray_group = dset = None
@@ -214,7 +215,8 @@ class VariableData:
 
                 if type_hint is DataType.PlotlyFigure:
                     import plotly.io as pio
-                    return pio.from_json(value.tobytes())
+                    b = value.tobytes()
+                    return pio.from_json(b) if deserialize_plotly else b.decode()
                 else:
                     return value
 
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -9,6 +9,7 @@ import h5py
 import numpy as np
 
 from .backend.db import BlobTypes, DamnitDB, blob2complex
+from .util import isinstance_no_import
 
 
 # This is a copy of damnit.ctxsupport.ctxrunner.DataType, purely so that we can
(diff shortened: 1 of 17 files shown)


damnit/backend/extraction_control.py
```diff
@@ -157,6 +157,8 @@ def _filter_env(self):
             env.pop(v, None)
         # Slurm jobs shouldn't attempt to use your X session
         env.pop('DISPLAY', None)
+        # Some LD_PRELOAD entries from FastX (?) cause spurious warnings
+        env.pop('LD_PRELOAD', None)
```
takluyver, line 161: This gets rid of warnings like this in the processing logs:

```
ERROR: ld.so: object 'libdlfaker.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
```

damnit/api.py
```diff
@@ -171,6 +171,56 @@ def summary(self):
                 return blob2complex(value)
             return value
 
+    def preview(self):
```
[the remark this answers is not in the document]
tmichela: *thinking out loud* I am not sure I like trying to infer the right data format at the API level. What about just returning whatever is saved in `.preview` or None if the preview is missing, and let front-end applications deal with handling alternatives?

tmichela (commented):
There 's also @CammilleCC 's questions about saving downsampled data rather than pngs for thumbnails. But that can maybe be a separate PR if we go that direction.

damnit/api.py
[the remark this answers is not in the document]
tmichela: >  I don't see any particular advantage to returning... an h5py group or dataset object? and do this logic in the application.

Ah, I think I didn't described accurately what I meant: 
I don't mean to return a hdf5 group/dataset, but in your current implementation you do 2 things that tried to infer what the data *should* look like:
- you check that the `preview` data is the right type/shape
- you fall back to read the data, and check that it's also the right type/shape

What I think is that this should only return what is saved to `.preview/` (or None if there's nothing), and that's up to the client to:
- do the check it it is able to display it
- get the data elsewhere and fallback if `preview` is not available

takluyver:
Aha, sorry, that makes more sense. I don't think we currently expose ways
to check things like the data shape and dtype before loading it, though, so
working out that it's not suitable for displaying a preview could be
costly.

We could do something similar when writing the file: if preview isn't
passed but data is a suitable type for preview, make a link in the hdf5
file pointing to the data object. Then we wouldn't need a fallback when
finding the preview. I don't have a strong preference on that.

tmichela:
>  I don't think we currently expose ways to check things like the data shape and dtype before loading it

That's' a good point, maybe that's the best approach to this problem?

takluyver:
I still think that there should be an automatic fallback to the data if no preview is provided and the data is usable as its own preview, either when accessing it (as currently implemented) or when writing the results in the HDF5 file. If you think doing the fallback at write time is preferable, I can do that.

I'm not against adding properties to check e.g. shape & dtype without loading the full data as well, but I think that's orthogonal to this PR, and it's not going to be absolutely trivial, with the xarray types involved.

takluyver:
I just realised, doing the fallback this way has the advantage that it immediately works for existing DAMNIT HDF5 output where we can display the data. Doing it at write time, we would either need to go and add `.preview` to all the existing files, or have a second fallback for older files.

tmichela:
> If you think doing the fallback at write time is preferable, I can do that.

I honestly like this even less :grimacing: 

And I still think that the API is not the place to decide what is displayable data or not.
My reasoning is that we currently have 2 clients, there maybe more if users want to have their own thing, and what a client can display does not necessarily match the others. So trying to return data only if it the right shape and dtype is something I don't particularly like, at least if it's saved in the .preview.

But maybe that's just me trying to project that future GUI capabilities will diverge *soon*. Any opinion on that, @JamesWrigley @CammilleCC ?

takluyver:
If you explicitly specify preview data in a context file, I think it should reject unsuitable objects - e.g. if you confuse preview & summary. That's my starting point. So that means that there needs to be a general definition of what is suitable as a preview, not specific to an application. This definition feels fairly natural to me: figure objects are designed to be displayed, and arrays up to 2 logical dimensions. The fallback to data if it meets those requirements then seems like the obvious thing to do both in isolation, and to fit with how DAMNIT has worked up until now.

Nothing prevents us expanding the definition of preview information going forwards, e.g. if we decide to support plots from another plotting library. Any application can support a subset of preview data types, refusing to display things it doesn't recognise. Specialised applications could also display data that we don't consider preview-compatible from the main data section.

takluyver:
I've moved the code to load the preview data out of the API and into the application layer.

damnit/api.py
```diff
@@ -171,6 +173,110 @@ def summary(self):
                 return blob2complex(value)
             return value
 
+    def preview_data(self, *, data_fallback=True, deserialize_plotly=True):
+        """Get the preview data for the variable
+
+        May return a 1D or 2D data array, a 3D RGB(A) arrray, a plotly figure
+        object, a str of plotly JSON (with deserialize_plotly=False) or None
```
JamesWrigley, line 180: Very nitpicky, sorry:
```suggestion
        May return a 1D or 2D data array, a 3D RGB(A) arrray, a Plotly figure
        object, a str of Plotly JSON (with deserialize_plotly=False) or None
```

damnit/ctxsupport/ctxrunner.py
```diff
@@ -465,6 +465,27 @@ def generate_thumbnail(image):
...
+    # width = 3 * height; roughly fits table cells
+    fig = Figure(figsize=(2, 2/3))
+    ax = fig.add_subplot()
+    fig.subplots_adjust(left=0, right=1, bottom=0, top=1)
+
+    if isinstance(arr, np.ndarray):
+        ax.plot(arr)
+    else:
+        # Use DataArray's own plotting method
+        arr.plot(ax=ax)
```
JamesWrigley, line 480: Is this necessary? I think for 1D plots the only extra thing xarray does is add the right axis labels and ticks, but those are disabled in the preview.

damnit/api.py
```diff
@@ -171,6 +173,110 @@ def summary(self):
...
+
+        return None
+
+    def preview(self):
+        """Show the preview data
+
+        This is intended for use in Jupyter notebooks
+        """
+        if (obj := self.preview_data()) is None:
+            return
```
JamesWrigley, line 243: Could we return a string like "Preview could not be generated"? That would be a bit clearer than the function seeming to do nothing when called.

damnit/api.py
```diff
@@ -171,6 +173,110 @@ def summary(self):
...
+                else:
+                    dset = grp['data']
+            else:
+                return None
+
+            if xarray_group is not None:
+                for obj in f[xarray_group].values():
+                    if isinstance(obj, h5py.Dataset) and (
+                        obj.ndim > 3 or (obj.ndim == 3 and obj.shape[-1] not in (3, 4))
+                    ):
```
JamesWrigley, line 213: A Dataset doesn't have `ndim` or `shape` properties, it's just a container for other arrays. We could perhaps return a string like the summary instead? Though that's not really a preview.

```suggestion
                    if isinstance(obj, h5py.Dataset):
```

JamesWrigley (changes requested):

damnit/api.py
JamesWrigley: Oh NVM, I misread this as an `xr.Dataset` rather than an `h5py.Dataset` :upside_down_face:

damnit/ctxsupport/ctxrunner.py
takluyver: I think it will use the coordinate labels as the x points, so if they're not evenly spaced - e.g. if you've done a scan with varying step sizes - this could make a difference. Probably not common, but I don't think it does any harm.

damnit/api.py
takluyver: I agree it makes sense to have something visible in that case; what do you think about throwing an error rather than returning a string?

JamesWrigley: Yep that seems fine.

JamesWrigley (approved):
LGTM!

PR #434, opened 2025-04-08
feat: LogViewWindow live updates

tmichela:
Poll the log file changes monitored by the LogViewWindow

### DIFF
--- a/damnit/gui/main_window.py
+++ b/damnit/gui/main_window.py
@@ -163,7 +163,7 @@ class MainWindow(QtWidgets.QMainWindow):
     def _create_status_bar(self) -> None:
         self._status_bar = QtWidgets.QStatusBar()
 
-        self._status_bar.messageChanged.connect(lambda m: self.show_default_status_message() if m == "" else m)
+        self._status_bar.messageChanged.connect(self.on_status_message_changed)
 
         self._status_bar.setStyleSheet("QStatusBar::item {border: None;}")
         self._status_bar.showMessage("Autoconfigure AMORE.")
@@ -171,6 +171,10 @@ class MainWindow(QtWidgets.QMainWindow):
 
         self._status_bar_connection_status = QtWidgets.QLabel()
         self._status_bar.addPermanentWidget(self._status_bar_connection_status)
+
+    def on_status_message_changed(self, msg):
+        if msg == "":
+            self.show_default_status_message()
 
     def show_status_message(self, message, timeout = 0, stylesheet = ''):
         if isinstance(stylesheet, StatusbarStylesheet):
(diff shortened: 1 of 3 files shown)


damnit/gui/main_window.py
```diff
@@ -1094,17 +1095,78 @@ def drawControl(self, element, option, painter, widget=None):
...
+
+    def _check_log_file_update(self):
+        """Poll the file for changes in mtime or size."""
+        try:
+            stat_res = self.file_path.stat()
+        except FileNotFoundError:
+            # File might have been deleted
+            if self._last_size is not None:  # Avoid repeated logging
+                log.warning(f"Log file {self.file_path} not found.", exc_info=True)
+                self.text_edit.setPlainText(f"[Log file {self.file_path} not found or deleted]")
```
takluyver, line 1129: I might just leave the previous contents visible in this scernario, but I don't feel strongly about it.

tests/test_gui.py
```diff
@@ -1405,3 +1405,73 @@ def test_filter_header(mock_db_with_data, qtbot):
...
+    assert initial_stat.st_size > 0
+    # Check internal state matches
+    assert window._last_size == initial_stat.st_size
+    assert window._last_mtime == initial_stat.st_mtime
+
+    with log_file_path.open("a") as f:
+        f.write(appended_text)
+
+    # Wait until the filesystem reflects the change
+    qtbot.waitUntil(lambda: log_file_path.stat().st_mtime > initial_stat.st_mtime, timeout=2000)
```
takluyver, line 1431: Is this wait necessary? I'd assume that changes made on the same node, let alone in the same process, are visible as soon as the file is flushed/closed above.

takluyver:
LGTM

(I can't stop myself mentioning that it could continue to show some outdated data if you delete/truncate a log file and then write more data than before within the polling interval. But I don't think that's likely to happen often, and the benefits of handling the normal case of log files growing efficiently seems more valuable.)

tests/test_gui.py
tmichela: you're right it's probably not necessary, I removed it.

tmichela:
Any idea what causes the tests to fail? The additional test should not even start a MainWindow, so it must be happening elsewhere 🤔

JamesWrigley:
The tests are somewhat segfaulty... I have some fixes in #394, you could try cherry-picking those to see if it helps:
- https://github.com/European-XFEL/DAMNIT/pull/394/commits/725492fc2d7e2a1f14a4cdeb750b2a20eedb550d
- https://github.com/European-XFEL/DAMNIT/pull/394/commits/78854f3d59a1f70e7a0fbab001062694bedf9a97
- https://github.com/European-XFEL/DAMNIT/pull/394/commits/7bb30457e6b2e0ae254ac2056b6cf2a8bfc977cd

tmichela:
> I have some fixes in https://github.com/European-XFEL/DAMNIT/pull/394, you could try cherry-picking those to see if it helps:

Looks like it fixed it, thanks!

PR #436, opened 2025-04-24
fix: do not store error for transient variables

tmichela:
This caused saving rows in the database for those transient vars leading the table to show extra columns with an integer index for those transient vars:

image

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -337,6 +337,13 @@ class ContextFile:
 
                 if missing_deps:
                     log.warning(f"Skipping {name} because of missing dependencies: {', '.join(missing_deps)}")
+                    # get error message from transient dependencies
+                    error_message = ''
+                    for dep in missing_deps:
+                        if self.vars[dep].transient and dep in errors:
+                            error_message += f'\ndependency ({dep}) failed: {repr(errors[f"{dep}"])}'
+                    if error_message:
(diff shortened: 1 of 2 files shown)


damnit/ctxsupport/ctxrunner.py
```diff
@@ -337,6 +337,13 @@ def execute(self, run_data, run_number, proposal, input_vars) -> 'Results':
 
                 if missing_deps:
                     log.warning(f"Skipping {name} because of missing dependencies: {', '.join(missing_deps)}")
+                    # get error message from transient dependencies
+                    error_message = ''
+                    for dep in missing_deps:
+                        if self.vars[dep].transient and dep in errors:
+                            error_message += f'\ndependency ({dep}) failed: {repr(errors[f"{dep}"])}'
+                    if error_message:
+                        errors[name] = Exception(error_message)
```
takluyver, line 346: I think the downstream dependent variables in this case should get a grey box, so it's easy to spot which ones are actually failing (orange boxes) from the ones that are skipped as a consequence. We can achieve this by using the `Skip` exception class, or we could introduce a new exception class and tweak the logic in the GUI to mark errors.

takluyver: Ah, sorry, I was reading too fast, this is *only* collecting errors from transient variables. I thought it would propagate errors from visible dependencies too.

tmichela:
Thanks!

PR #438, opened 2025-04-28
fix: prevent the gui to crash if there are no selected rows in the table

tmichela:
@philsmt reported this in zulip some while ago, and it has happened to me this week as well.
So we just catch the error here to prevent the gui to crash when we right click on a row (to reprocess/view logs).

### DIFF
--- a/damnit/gui/table.py
+++ b/damnit/gui/table.py
@@ -322,7 +322,11 @@ class TableView(QtWidgets.QTableView):
         try:
             row = self.selected_rows()[0].row()
         except IndexError:
-            # not sure how this happens but it does
+            self.damnit_model._main_window.show_status_message(
+                "No row selected",
+                timeout=7000,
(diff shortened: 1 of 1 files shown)


takluyver:
Is it worth making a somewhat visible error when it occurs, maybe in the status bar? I'd personally rather see that something went wrong than just have the application silently ignore me. I don't feel strongly about this, though, so if you want to go for it as is, LGTM.

If this comes up often enough to be annoying even when it doesn't crash, we could use the table view's `.itemAt` or `.rowAt` method on right-click to identify the row you're right clicking on, rather than relying on that being selected.

takluyver:
I've opened #439 to fix the same problem, but I think we can land this one as well.

tmichela:
I added showing an error message, thought this is probably not needed anymore with #439

takluyver:
LGTM

PR #439, opened 2025-04-28
Only enable 'View processing logs' action when rows selected

takluyver:
Disable the context menu option when it can't work. I think making it visibly disabled is nicer UX, but we could merge #438 as well for safety, in case there's a corner case I've missed.

Screenshot From 2025-04-28 17-50-21

Screenshot From 2025-04-28 17-50-37


tmichela:
I notice we can also reproduce this with <CTRL> or <SHIFT> + right click on a row, this would open the menu but not select the row.

This works reliably, thanks! LGTM

tmichela (approved):

PR #445, opened 2025-05-04
Make `add_to_h5_file()` set the file permissions

JamesWrigley:
Otherwise if an exception is raised while writing the file would have the usual umask permissions set, which would block other users from overwriting the file.

This was affecting MID so I went ahead and deployed the branch.


takluyver:
LGTM

PR #447, opened 2025-05-07
Fix pre-creating runs in an empty table

takluyver:
This has come up a couple of times recently.


egorsobolev:
LGTM! Thanks, @takluyver !

PR #448, opened 2025-05-08
Allow variable functions to return Cell(data=None)

takluyver:
`return Cell(None)` caused a crash when saving the results. We decided this should be allowed, and it should be possible to store preview & summary without actual data. This needed a tweak to the API so it can still access such variables: `.read()` will get the summary value, like it does for user-editable variables.

For executing the context file, `return Cell(None)` has the same effect as `return None`: dependent variables will either be skipped, or use a default value if one is given.

### DIFF
--- a/damnit/api.py
+++ b/damnit/api.py
@@ -324,7 +324,7 @@ class RunVariables:
         titles_to_names = { title: name for name, title in names_to_titles.items() }
 
         if name not in key_locs and name not in titles_to_names:
-            raise KeyError(f"Variable data for '{name!r}' not found for p{self.proposal}, r{self.run}")
+            raise KeyError(f"Variable data for {name!r} not found for p{self.proposal}, r{self.run}")
 
         if name in titles_to_names:
             name = titles_to_names[name]
@@ -339,6 +339,10 @@ class RunVariables:
         with h5py.File(self.file) as f:
             all_keys = { name: False for name in f.keys()
                          if not name.startswith('.')}
+
+            # Add keys with summary but not data in file
+            all_keys.update({ name: False for name in f['.reduced'].keys()
+                              if name not in all_keys})
 
         # And the keys from the database
         user_vars = list(self._db.get_user_variables().keys())
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -283,7 +283,8 @@ class ContextFile:
         return ContextFile(new_vars, self.code)
 
     def execute(self, run_data, run_number, proposal, input_vars) -> 'Results':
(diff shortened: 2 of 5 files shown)


damnit/api.py
```diff
@@ -122,7 +122,7 @@ def read(self, deserialize_plotly=True):
             deserialize_plotly (bool): Whether to deserialize Plotly figures
                 into `Figure` objects. If this is `False` the JSON string will be returned.
         """
-        if self._db_only:
+        if self._summary_only:
```
tmichela, line 125: :thinking:  I would probably not want to fall back to summary value in the case where the data isn't present. My thinking is, it should be as equally valid to not have summary value saved. e.g. you save preview data and some background color. I can also imagine support for more things in Cell in future.

I'm thinking user variables is a special case which has it's actual data saved in the database (as summary) and nothing in the h5 file.

takluyver: Fair enough, so back to `db_only` and let `.read()` fail for computed variables without data?

Should these still show up in `damnit[52].keys()` even if there's no data to read?

FYI, I think it's currently impossible to have preview with no summary, because one of the fallback summary options will supply one. Of course that might change in the future.

tmichela: > FYI, I think it's currently impossible to have preview with no summary, because one of the fallback summary options will supply one. Of course that might change in the future.

Oh, OK :eyes: so yea, I would then definitely not fall back to the summary value, because in such case I don't think you'd ever want to reuse the summary data.

> Fair enough, so back to db_only and let .read() fail for computed variables without data?
Should these still show up in damnit[52].keys() even if there's no data to read?

Hum, I personally would prefer that any existing variable existing in the table is always visible in `keys` even if there's no data for this var in this run, and it would return None in such case rather than fail.
I guess that would break the current behavior (?).
I haven't used the API much so far but our users starts to do so, and generally I prefer testing for data than catching errors.
@JamesWrigley you probably have an opinion here? Since you've likely used the most the API so far.

takluyver: Yeah, the current design is that `damnit[52].keys()` gives you only the set of variables which have data for that run. As this PR stands it would extend it to include variables which returned a `Cell()` even if there's no `data` part.

JamesWrigley: My preference would be to always throw an error because I think that's clearer/easier to debug than `.read()` returning `None` and some later code failing with a type error.

takluyver: As the function stands it will throw a `KeyError` from h5py when it tries to look up the name:

```text
KeyError: "Unable to synchronously open object (object 'foo' doesn't exist)"
```

That seems fairly clear, but we could catch it & re-raise with a more specific message if you'd like.

JamesWrigley: Nah I think that's fine :+1:

takluyver:
Thanks!

PR #452, opened 2025-05-15
Fix icon lookup path

JamesWrigley:
This probably broke when the function was moved in a recent refactor.


takluyver:
Thanks, that was me. :sweat_smile: LGTM

PR #453, opened 2025-05-15
Fix ctxrunner compatibility with Python 3.9

JamesWrigley:
The previous style only works on 3.10+:
https://discuss.python.org/t/call-staticmethod-from-class-body/14542/2

Also renamed the `_cache` dict to avoid confusion with the `_cache()` decorator.

Not sure how far we want to take backwards compatibility, but this was an easy fix so :shrug: It came up in the wild from an old proposal using a custom Python 3.9 environment.


takluyver:
LGTM

If it's important to keep it working on a 3.9 environment, we should maybe figure some way of testing for that. I also don't know how far we need to take it, though. I don't think long-term reproducibility is really a priority for DAMNIT, although it's always nice.

JamesWrigley:
Agreed, if it comes up again I'll add a test for Python 3.9 support.

PR #455, opened 2025-05-22
Flush stdout after each variable function

takluyver:
When we use print() in context files for debugging, the printed output can appear after the log message that that variable has finished, which is confusing. This is because logs go to stderr, and stdout & stderr have different [buffering rules](https://docs.python.org/3/library/sys.html#sys.stdout).

This should resolve that by flushing stdout after the variable function finishes, before we log anything.


JamesWrigley (approved):
LGTM!

PR #475, opened 2025-07-09
Option to copy user-editable vars from proposal

takluyver:
@CammilleCC mentioned that SCS often copy context files from previous proposals, but it's an issue that user-editable variables aren't copied as well. This adds an option for that, selected by default if you're copying from a previous proposal.

image


JamesWrigley (changes requested):
Gotta have at least one test :sweat_smile: Otherwise it looks good.

JamesWrigley (approved):
LGTM!

PR #476, opened 2025-07-09
Update the default `context_python` for the 202502 environment

JamesWrigley:
Tested by changing the environment of a database and reprocessing some variables.


takluyver:
LGTM!

PR #477, opened 2025-07-09
Fix some warnings about pandas deprecations

takluyver:
- `df.applymap()` was renamed to `.map()` in pandas 2.1.
- The `make_finite` function raised a warning about `.fillna()` downcasting from object dtype. It's not very clear what this is doing - it doesn't get rid of infinities in the data - and it's now only used in one place, so I've simplified the code a bit.


JamesWrigley (approved):
Perhaps the best way to see what `make_finite()` is doing is to remove it and see if anything breaks :upside_down_face: I can't think of anything but we can add it back later if there's a need. LGTM!

PR #480, opened 2025-07-24
feat: dependency pattern matching annotation

tmichela:
Allow unix shell style pattern matching in variable dependency annotation to match multiple deps, e.g.

```python3
@Variable()
def my_var(run, data: 'var#xgm_[123]'):
    # data is a dict[str, Any] where the keys are dependency names
    return sum(data.values())
```

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -6,7 +6,7 @@ possibly running in a different Python interpreter. It will run a context file
 """
 
 import argparse
-import functools
+import fnmatch
 import inspect
 import io
 import logging
@@ -191,14 +191,25 @@ class ContextFile:
         if problems:
             raise ContextFileErrors(problems)
 
-    def ordered_vars(self):
+    def direct_dependencies(self, *variables: Variable) -> set[str]:
+        """return a set of names of direct dependencies of the passed Variable's
+        """
+        dependencies = set()
+        for variable in variables:
+            for dependency in variable.arg_dependencies().values():
+                # expand matching patterns to match all variable dependencies
+                dependencies.update(fnmatch.filter(self.vars, dependency))
+        return dependencies
+
+    def ordered_vars(self) -> tuple[str]:
         """
         Return a tuple of variables in the context file, topologically sorted.
         """
-        vars_graph = { name: set(var.arg_dependencies().values()) for name, var in self.vars.items() }
+        ts = TopologicalSorter()
 
-        # Sort and return
-        ts = TopologicalSorter(vars_graph)
+        for name, var in self.vars.items():
+            ts.add(name, *self.direct_dependencies(var))
+
         return tuple(ts.static_order())
 
     def all_dependencies(self, *variables):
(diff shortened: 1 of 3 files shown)


damnit/ctxsupport/ctxrunner.py
```diff
@@ -304,9 +315,14 @@ def execute(self, run_data, run_number, proposal, input_vars) -> 'Results':
...
                         dep_name = annotation.removeprefix("var#")
-                        if dep_name in dep_results:
+                        match = fnmatch.filter(dep_results, dep_name)
+
+                        if len(match) == 1 and match[0] == dep_name:
                             dep_data = dep_results[dep_name]
                             kwargs[arg_name] = dep_data
+                        elif len(match) > 1:
+                            dep_data = {name: dep_results[name] for name in match}
+                            kwargs[arg_name] = dep_data
```
takluyver, line 325: While we're at it we can make this code slightly shorter:

```suggestion
                            kwargs[arg_name] = dep_results[dep_name]
                        elif len(match) > 1:
                            kwargs[arg_name] = {name: dep_results[name] for name in match}
```

damnit/ctxsupport/ctxrunner.py
```diff
@@ -304,9 +315,14 @@ def execute(self, run_data, run_number, proposal, input_vars) -> 'Results':
                     # Dependency within the context file
                     if annotation.startswith("var#"):
                         dep_name = annotation.removeprefix("var#")
-                        if dep_name in dep_results:
+                        match = fnmatch.filter(dep_results, dep_name)
+
+                        if len(match) == 1 and match[0] == dep_name:
                             dep_data = dep_results[dep_name]
                             kwargs[arg_name] = dep_data
+                        elif len(match) > 1:
```
takluyver, line 323: So a wildcard dependency implies we'll skip running the downstream variable, it will never be passed an empty dict? I'm OK with that, as you can always give it a default `{}` if you want it to run with no matches.

docs/backend.md
```diff
@@ -111,6 +111,25 @@ def bar(run, value: "var#foo"):
...
+@Variable()
+def base_1(run):
+    return 1
+
+@Variable()
+def base_2(run):
+    return 2
+
+@Variable()
+def sum_values(run, data: 'var#base_?'):
```
takluyver, line 128: Small nitpick: I might illustrate the wildcards with `*`, because I think more people will recognise that as a familiar mechanism. I know what `?` means, but it's much rarer, so I assume a lot of people don't know it's part of the same system.

docs/backend.md
```diff
@@ -111,6 +111,25 @@ def bar(run, value: "var#foo"):
     return value * 2
 ```
 
+Pattern matching using unix shell-style
+[wildcards](https://docs.python.org/3/library/fnmatch.html) is allowed in `var#`
```
takluyver, line 115: We refer to these as glob patterns [in the EXtra-data docs](https://extra-data.readthedocs.io/en/latest/reading_files.html#extra_data.DataCollection.select) (along with a brief description). I don't feel strongly about this, though.

docs/backend.md
tmichela: updated in 847df374ee47f675fc6a70b8ed26beea4658badb

docs/backend.md
tmichela: sure, I just took the mention as it was from the fnmatch docs, did not know of the reference in extra-data.

updated in 847df374ee47f675fc6a70b8ed26beea4658badb

tmichela:
thanks!

PR #482, opened 2025-08-04
Docs: Update module name to damnit

takluyver:
I've just created these new names as copies of the amore modules.


JamesWrigley (approved):
LGTM!

PR #485, opened 2025-08-08
Use Python 3.13 in CI

JamesWrigley:
To match the deployed version. Though maybe in the future we should change this to use pixi?


JamesWrigley:
I'll close this in favour of #486.

PR #486, opened 2025-08-11
Attempt to use Pixi to run tests in CI

takluyver:


pixi.lock
```diff
@@ -407,7 +464,7 @@ packages:
 - pypi: ./
   name: damnit
   version: 0.2.0
-  sha256: f6fca7a312f9fbdc2b30e10f56e2cdd46af8660a5fc8cb917a93db97d35296bf
+  sha256: 1984fbc62a966502b0744f685bad527d7a6fa1c726be36222a6ca1bc19d22fa4
```
takluyver, line 467: It looks like we will need to update the lock file whenever `pyproject.toml` changes - this is the hash of that file. Not ideal, but I suggest we go with it for now and see how it works in practice.

JamesWrigley: Could we add it to the [release docs](https://damnit.readthedocs.io/en/latest/internals/#making-a-new-release)?

PR #509, opened 2025-10-15
refactor: use extra-proposal to query mymdc

tmichela:


damnit/ctxsupport/ctxrunner.py
```diff
@@ -336,10 +258,10 @@ def execute(self, run_data, run_number, proposal, input_vars) -> 'Results':
...
                     # Mymdc fields
                     elif annotation.startswith("mymdc#"):
-                        if mymdc is None:
-                            mymdc = MyMetadataClient(proposal)
-                        metadata = annotation.removeprefix('mymdc#')
-                        kwargs[arg_name] = getattr(mymdc, metadata)(run_number)
+                        if metadata is None:
+                            metadata = extra_proposal.Proposal(proposal)[run_number]
+                        field = annotation.removeprefix('mymdc#')
+                        kwargs[arg_name] = getattr(metadata, field)()
```
JamesWrigley: Not sure how I feel about this in hindsight, it does feel a bit weird to allow calling arbitrary functions. Shall we restrict it to supported fields?

JamesWrigley (changes requested):

damnit/ctxsupport/ctxrunner.py
tmichela: It's already enforced by the [check](https://github.com/European-XFEL/DAMNIT/blob/master/damnit/ctxsupport/ctxrunner.py#L184) step when loading the context file.
I can make it explicit here as well if you feel strong about it.

docs/configuration_and_management.md
```diff
@@ -252,6 +252,7 @@ $ damnit db-config context_python /path/to/your/python
 The environment *must* have these dependencies installed for DAMNIT to work:
 
 - `extra_data`
+- `extra_proposal`
 - `pyyaml`
 - `requests`
```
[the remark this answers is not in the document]
tmichela: updated in 19bfb8d4820ac369f77ba42bc3415b7805a0bfe1

JamesWrigley (approved):
LGTM!

PR #510, opened 2025-10-20
Feat: migrations framework

tmichela:
Separates the database migration code from the DamniDB class, should be easier to implement database schema updates.

### DIFF
--- a/damnit/backend/db.py
+++ b/damnit/backend/db.py
@@ -130,11 +130,18 @@ class DamnitDB:
         return self._path
 
     def _set_schema_version(self, version: int):
-        self.metameta["data_format_version"] = int(version)
+        self.metameta["data_format_version"] = version
 
-    def upgrade_schema(self, from_version):
+    def upgrade_schema(self, from_version: int):
         to_version = latest_version()
-        log.info("Upgrading database format from v%d to v%d", from_version, to_version)
+
+        if from_version == 0:
+            raise RuntimeError(
+                "Database is at legacy format v0, please contact da-support@xfel.eu for help migrating")
+        elif from_version < 0:
+            log.info("Creating new database at %s with format v%d", self._path, to_version)
+        else:
+            log.info("Upgrading database format from v%d to v%d", from_version, to_version)
 
         applied = apply_migrations(
             self.conn,


takluyver:
I think we're going to need to revisit the 'automatically migrate on open' mechanism, because old DAMNIT folders become read-only, and the migration will fail. We could:

1. Abstract over the differences in the code reading the DB, so it can read old databases without migrating them. The neatest approach in some ways, but it would require a fair bit of ongoing work - every query we write, we have to think about whether it works with 4 different schemas.
2. If the DB is read-only, copy it to a temp file, upgrade that and use it instead. Simple, but overhead every time you open it.
3. If the DB is read-only, copy it to a predictable, stable location (maybe in proposal scratch?), upgrade and use that. If it's opened again, the predictable location should be found and used instead - effectively a cache of the upgraded read-only DB. We'd still want to track that it's conceptually read-only, so you can't change the data in it.

damnit/backend/db_migrations.py
```diff
@@ -0,0 +1,181 @@
...
+    set_version: Callable[[int], None],
+) -> list[Migration]:
+    """
+    Apply all migration steps to go from `from_version` up to `to_version`.
+    Sets the version after each successful step using `set_version`.
+    Returns the list of applied migrations.
+    """
+    applied: list[Migration] = []
+    for step in pending_migrations(from_version, to_version):
+        # Each step in its own transaction for clarity and resilience
```
takluyver: I'm not sure what's common for these things, but I'd be inclined to put them all in 1 transaction. If I try to migrate a v1 database to v4 and it errors out, I'd prefer to be left with the unchanged v1 database rather than a v2 or v3 database. 🤷

damnit/backend/db_migrations.py
```diff
@@ -0,0 +1,181 @@
...
+    - `description`: short human description for docs/logs
+    - `apply`: a callable that takes a sqlite3.Connection and applies changes
+    """
+
+    to_version: int
+    description: str
+    apply: Callable[[sqlite3.Connection], None]
+
+
+def _to_v1(conn: sqlite3.Connection) -> None:
```
takluyver: Purely stylistic, but it would be neat to associate the numbers and the descriptions more closely with the functions. Maybe via a decorator:

```python
MIGRATIONS = []

def migration_func(to_version):
    def inner(func):
        MIGRATIONS.append(Migration(to_version, func.__doc__, func))
    return inner

@migration_func(1)
def _to_v1(conn: sqlite3.Connection):
    """Baseline schema (core tables, views)"""
    ...
```

damnit/backend/db_migrations.py
```diff
@@ -0,0 +1,181 @@
...
+    conn.executescript(
+        dedent(
+            """
+            -- Core schema (baseline)
```
takluyver: Further stylistic comment: a lot of this file ends up 12 spaces indented. I think it's a good idea to keep the flexibility of a callable, not just an SQL script, but I wonder if it's worth a shortcut wrapping a string for the cases where it is just SQL?

Alternatively, maybe we just reformat it to knock off some whitespace:


```suggestion
    conn.executescript(dedent(
        """
        -- Core schema (baseline)
```

damnit/backend/db.py
```diff
@@ -185,50 +129,25 @@ def kafka_topic(self):
     def path(self):
         return self._path
 
+    def _set_schema_version(self, version: int):
+        self.metameta["data_format_version"] = int(version)
```
JamesWrigley: Is the explicit conversion necessary since the type hint says `int`? If so then the type hint should probably be changed.
```suggestion
        self.metameta["data_format_version"] = version
```

damnit/backend/db_migrations.py
```diff
@@ -0,0 +1,226 @@
...
+    to_version: int
+    description: str
+    apply: Callable[[sqlite3.Connection], None]
+
+
+MIGRATIONS: list[Migration] = []
+
+
+def migration_step(
+    to_version: int, *, schema: str | None = None, description: str = ""
```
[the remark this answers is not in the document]
tmichela: They're equivalent, but tools like mypy suggest using None is [better](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type).

JamesWrigley:
Me, I regularly git blame :sob:

PR #513, opened 2025-10-24
fix: allow preview data for boolean arrays

tmichela:


tmichela:
I'll merge this end of this week if there's no objections.

PR #515, opened 2025-10-30
Fix support for opening readonly databases

JamesWrigley:
Previously opening old readonly databases would fail because DamnitDB would try to `.setdefault()` some settings, now those are only set either when the database is created or when actually required for reprocessing (in which case write permissions are needed anyway).
Also bumped the version because I'd like to make a new release and update our environments.

As an aside, I noticed when writing this that the stable module is still under the old `amore-mid` directory :sweat_smile:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,20 +11,10 @@ env:
 
 jobs:
   tests:
-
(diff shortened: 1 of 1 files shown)


takluyver:
LGTM

PR #517, opened 2025-11-06
fix: always execute the context file in a sandbox

tmichela:
Also:
- fixes a few typos/bugs I found from the central listener update.
- always execute the context file in a subprocess.


JamesWrigley:
Fair, as the author I'm probably a bit biased there :trollface:

JamesWrigley (approved):
Lazy LGTM! 🦥

PR #521, opened 2026-01-16
Add support for `--no-kafka` to `damnit read-context`

JamesWrigley:
This is also supported by the GUI now, so if the GUI is started with `--no-kafka` the command it spawns will pass that as well.


tmichela (approved):
LGTM

PR #525, opened 2026-01-23
Rename 'GUI' to 'Getting started' and remove old web section

JamesWrigley:
Not sure how I missed that there was already a section about the web interface :facepalm: It was out of date anyway so I just deleted it.

The rename is because 'getting started' feels more descriptive of the content, and because people will likely recognize that phrase more than just 'GUI'. Renaming the file itself does have the downside of breaking links but I don't think that matters very much for this.


tmichela (approved):
LGTM

PR #526, opened 2026-01-27
feat(gui): save column width

### LINKED ISSUE #258: Save column width
When closing and reopening the GUI, besides the order of the columns also their displayed/adjusted width should be recovered. 

Currently this is not the case which is particularly annoying for tables with many columns.

tmichela:
Adding long requested saving of column width information to local store.


JamesWrigley (changes requested):

JamesWrigley (changes requested):

JamesWrigley (approved):
LGTM!

PR #529, opened 2026-01-30
Ignore sqlite ResourceWarnings in pytest

takluyver:


takluyver:
This has hidden most of the warnings, but a few are still visible between the last test and the 'warnings summary' line. I tried adding the same filter with a `-W` option in the python command pixi is configured to run, but it made no difference, so I removed it again.

I guess pytest changes the warning filters somehow as tests run. The warnings occur during garbage collection, so they get associated with essentially random lines of code depending on where garbage collection is triggered, making it hard to filter them predictably. It might be possible to control them a bit better by running `gc.collect()` after each test. :shrug:

tmichela:
LGTM

PR #535, opened 2026-02-25
Switch Kafka library back to kafka-python

takluyver:
It seems that the original `kafka-python` package, which we used until #283, is now maintained again, and kafka-python-ng has been archived.

https://github.com/kafka-python-ng/kafka-python-ng/issues/210


JamesWrigley (approved):
LGTM!

PR #538, opened 2026-03-02
Tests: send updates to a simplified Kafka broker instead of mocking

takluyver:
Instead of mocking the `KafkaProducer` & `KafkaConsumer` Python objects, this runs a mock Kafka broker in a background thread, and the real client objects send messages to it.

### DIFF
--- a/damnit/backend/combine.py
+++ b/damnit/backend/combine.py
@@ -14,7 +14,7 @@ from kafka import KafkaConsumer, KafkaProducer
 from xarray.backends import H5NetCDFStore
 
 from ..context import DataType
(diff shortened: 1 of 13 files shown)


damnit/backend/combine.py
```diff
@@ -97,19 +98,25 @@ def shutdown(self):
...
-                        "Unexpected error processing file submission message",
-                        exc_info=True,
-                    )
+            try:
+                self.handle_one_message()
+            except StopIteration:
+                pass  # consumer timeout was reached
+
+    def handle_one_message(self):
+        record  = next(self.consumer)
```
tmichela, line 107:
```suggestion
        record = next(self.consumer)
```

tmichela (commented):
LGTM

PR #539, opened 2026-03-09
feat: Pipeline

tmichela:
Public API for loading, inspecting, and executing context files programmatically.


JamesWrigley:
My apologies for letting this languish for so long 🙈 I'll try to review it this week.

docs/configuration_and_management.md
```diff
@@ -512,6 +512,131 @@ class DetectorAnalysisAlt(BaseAnalysis):
...
+    return 123
+
+@Group
+class Demo:
+    @Variable
+    def val(self, run):
+        return 456
+
+pipe = Pipeline()
+pipe.add(my_var, Demo(name="demo"))
```
JamesWrigley, line 562: Just to confirm, will this also work for reusing variables defined in other libraries?

JamesWrigley: (discussed offline, the answer is yes)

docs/configuration_and_management.md
```diff
@@ -512,6 +512,131 @@ class DetectorAnalysisAlt(BaseAnalysis):
...
+def calibrate_data(run, run_nb: 'meta#run_number'):
+    return calibration(run, run_nb)
+
+
+@Variable
+def baseline(run, run_nb: 'meta#run_number', proposal: 'meta#proposal'):
+    dark_run = get_last_dark(run_nb)
+
+    pipe = Pipeline(proposal=proposal, run_number=dark_run)
+    pipe.add(calibrate_data)
```
JamesWrigley, line 630: This feels like a dodgy example to me because one could just read the value from the old run with the API. Is there a way to replace arguments to a variable? If so then calling the variable with different arguments would make sense. Though my preference in that case would be to just call `calibration()` directly 😛

JamesWrigley (approved):
Didn't have time to do a proper review of the code and likely won't for the next few weeks, alas. But semantically this looks fine and I trust that since it's used at HED there's no blocking bugs, so to avoid holding it up further, LGTM!

docs/configuration_and_management.md
tmichela: Yes, I've made the example minimalistic on purpose, to keep it short and only show the concept of it, but I agree this isn't a realistic example.
A real case would be a combination of multiple variables and modified inputs based on current run's conditions.

I'll just add a simple input in this example.

JamesWrigley: Ok, but no need to let it hold up merging.

PR #540, opened 2026-03-11
Add support for returning matplotlib Axes as Cell previews

JamesWrigley:
We already support returning `Axes` directly, so it makes sense to also support them as previews.

CC @philsmt


tmichela (approved):
LGTM

PR #546, opened 2026-04-08
Update docs about listeners

takluyver:
Going forwards, we expect people to use the central listener by default, and static listeners as a special case. This reworks the description to reflect that.


JamesWrigley (approved):
LGTM!

PR #547, opened 2026-04-08
Feat(gui): Adds provenance markers

tmichela:
Adds a visual indicator now that we can edit the table from the API.

https://github.com/user-attachments/assets/4db68e67-19a1-40b8-a232-454aac230dcd


JamesWrigley (approved):
I'm now wondering if `source` is a more user-friendly term than `provenance` 🤔 But that should be changed in another schema change which is out of scope for this PR, so LGTM!

PR #548, opened 2026-04-08
Convert tags on group variable back to tuple

takluyver:
Because a set can't be serialised as JSON.

Fixes an issue sending updates to the frontend.


tmichela (approved):
LGTM!

takluyver:
Thanks!

PR #550, opened 2026-04-09
fix(plotly): inline plotly js

tmichela:


tmichela:
I'll merge this right away, as it's blocking Danilo.

PR #551, opened 2026-04-15
Show logs from the combiner service (in journal)

takluyver:
In #512 I missed the call to `logging.basicConfig()` (or something equivalent) to actually send the logs from the combiner anywhere. This caused some confusion when trying to debug it.

I've also added `SysLogIdentifier` options for systemd - this makes it easier to distinguish the services in the combined view that `damnit-logs` gives. Without it, the name used is the executable that's running, e.g. `damnit`.


JamesWrigley (approved):
LGTM!

Note that to deploy it you'll want to either rebase `feat/pipe` on top of master or switch to `master` in the beta module.

PR #552, opened 2026-04-16
Set umask to 0 in the ctxrunner process

JamesWrigley:
This ensures that extra files created by the context file (e.g. xwiz files) will by default not be blocked from setting 0o666 permissions.

CC @turkot


turkot:
Hi James, All,
This change is more urgent than we have initially thought - could it be deployed soon please?

takluyver:
LGTM

PR #555, opened 2026-04-17
Log timings for database write transactions

takluyver:
Borrowing a timing class we already use in calibration


takluyver:
I've just wrapped the transactions that the combiner uses for now, as we're trying to work that out. But we could wrap all write transactions like this.

tmichela:
LGTM

PR #556, opened 2026-04-19
Debugging stuff

JamesWrigley:
This is currently deployed. I've tried to keep the commits atomic so I'd recommend reviewing them one-by-one.

### DIFF
--- a/damnit/backend/db.py
+++ b/damnit/backend/db.py
@@ -257,7 +257,7 @@ class DamnitDB:
             for row in self.conn.execute("""
                 SELECT name, title, type, description, attributes FROM variables
                 WHERE type IS NULL
-            """):
+            """).fetchall():
                 var = dict(row)
                 vars_in_db[var.pop("name")] = var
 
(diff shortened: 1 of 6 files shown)


damnit/backend/db.py
```diff
@@ -89,6 +91,18 @@ def blob2numpy(data: bytes) -> np.ndarray:
...
+    tries = 10
+
+    for i in range(tries):
+        try:
+            return conn.execute(*args)
+        except sqlite3.OperationalError:
+            if i == tries - 1:
+                raise
+            else:
+                time.sleep(i * random.random())
```
takluyver: SQLite is already doing sleep-and-retry up to a configurable time limit (5 s by default) before we get the locked error:

https://sqlite.org/c3ref/busy_timeout.html

https://github.com/sqlite/sqlite/blob/63984b6ab762c97d6377013ab33b9116e12aa113/src/main.c#L1726-L1747

Do we have any particular reason to think that adding another layer of retries is more effective than just increasing the timeout and letting sqlite retry for longer?

If we do want to keep the extra layer of retrying, I would:

- Turn down the number of retries: with sqlite's own 5 second timeout times 10, plus around 20 seconds of total extra waits, this currently means well over a minute waiting if the database stays locked.
- Make it more specific: we can check `e.sqlite_errorcode == sqlite3.SQLITE_BUSY` to skip doing this for other types of error

takluyver (commented):
Other than a question about the retry mechanism, I think these changes looks fine, though I would be surprised if any of them made a dramatic difference.

damnit/backend/db.py
JamesWrigley: Hmm I see 🤔 I don't think I see a reason for this then, I'll just bump up the wait time to 10s.

JamesWrigley: Ah NVM, we already set it to 30s :upside_down_face:

PR #558, opened 2026-04-23
Fix(gui): redraw hierarchical header on horizontal scroll

tmichela:
before:

https://github.com/user-attachments/assets/3149ba52-42f8-43f0-b338-458975f62a70

after:

https://github.com/user-attachments/assets/44898214-0d92-4885-920c-f769e70351e2

### DIFF
--- a/damnit/gui/table.py
+++ b/damnit/gui/table.py
@@ -161,13 +161,20 @@ class FilterHeaderView(QtWidgets.QHeaderView):
                     continue
 
                 if is_group:
-                    self._paint_group_background(painter, visible_rect)
+                    self._paint_group_background(painter, group_rect, clip_rect=visible_rect)
 
-                text_rect = visible_rect.adjusted(4, 0, -4, 0)
+                text_rect = group_rect.adjusted(4, 0, -4, 0)
                 if text_rect.width() <= 0:
                     continue
 
-                painter.drawText(text_rect, Qt.AlignVCenter | Qt.AlignHCenter | Qt.TextWordWrap, text)
+                painter.save()
+                painter.setClipRect(visible_rect)
+                painter.drawText(
+                    text_rect,
+                    Qt.AlignVCenter | Qt.AlignHCenter | Qt.TextWordWrap,
+                    text,
+                )
+                painter.restore()
 
     def update_filtered_columns(self, filtered_cols):
         """Update the set of filtered columns and trigger repaint"""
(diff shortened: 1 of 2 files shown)


daniloefl:
This looks amazing. I find it hard to evaluate the code quality itself, but I tried it out in a use-case and I see that this mostly works well, but two columns (which are the ones that have been added later in the context file and DB) still seem to be messed up. The problem seem to be a matter of ordering of the columns, such that columns belonging to the same group are drawn consecutively.

[Screencast from 2026-04-23 16-30-39.webm](https://github.com/user-attachments/assets/8cab3d48-2df9-469d-9187-3a3ef155700d)

damnit/gui/table.py
daniloefl, line 1117: Perhaps sort columns by group here, if groups are used?

daniloefl (commented):
That is very good, I only propose to order columns based on whether they belong to a group early on.

damnit/gui/table.py
tmichela: good idea. changed it in b5eac67c2a379dfdddd8b71dcb5a1bc392f9dcb9

daniloefl: I tried it and there was no change. Is it sorting by group, or by title? Or is the group part of the title?

https://github.com/user-attachments/assets/c876a73d-a43d-4875-9ecd-083c66c8335d

takluyver: How does this interact with saving and restoring (below) the column order?

If we need to get rid of the saved column order, I'm fine with that - the saved order has always been something of a headache, and I think we have better ways of dealing with the number of columns now (especially using tags).

In any case, I suspect this PR is an improvement on the status quo, and we don't need to hold it up over this question.

takluyver:
LGTM as far as I can see - I still barely know the low-level painter APIs

damnit/gui/table.py
takluyver: (Also, if we do remove the saved order mechanism, I think we should arrange columns based on the order of variables in the context file, rather than alphabetically)

tmichela: > How does this interact with saving and restoring (below) the column order?

I made it such that the saved order takes precedence, so the current behavior isn't changed here.

> If we need to get rid of the saved column order, I'm fine with that

I'd be fine with that too

> we should arrange columns based on the order of variables in the context file

Probably yes.

daniloefl: For transparency: the reason why those columns are where they are above is because those variables were added a posteriori in my context file, and as a result, they are moved automatically to the end. I assume this happens elsewhere (when adding new columns to the DB?). If keeping the order is desired, this is fine, I suppose.

tmichela:
Thanks. I'll merge as is, to fix quickly the header rendering.
We can continue discussion offline about ordering and change that in another PR.

PR #559, opened 2026-04-24
Fix getting run start times in the GUI on update notification

takluyver:
Tested interactively on FXE's p11437.

While I was looking at this code, I also realised it could be simplified a bit, getting rid of passing around several dicts with similar keys.


JamesWrigley (approved):
LGTM!

PR #561, opened 2026-04-24
Count a variable as skipped if its dependencies are skipped

takluyver:
At present, a dependent variable is recorded as skipped if its dependencies return None, but as having an error if its dependencies raise `Skip`:

Screenshot From 2026-04-24 16-51-21

With this change, the `Skip` is propagated instead, making it easier to see which variables actually hit an error:

image

### DIFF
--- a/damnit/ctxsupport/ctxrunner.py
+++ b/damnit/ctxsupport/ctxrunner.py
@@ -515,7 +515,7 @@ class ContextFile:
                     log.warning(f"Skipping {name} because of missing dependencies: {', '.join(missing_deps)}")
                     # get error message from transient dependencies
                     dep_errors = [(dep, errors[dep]) for dep in missing_deps if dep in errors]
-                    if dep_errors:
+                    if dep_errors and not all(isinstance(e, Skip) for _, e in dep_errors):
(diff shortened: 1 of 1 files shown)


damnit/ctxsupport/ctxrunner.py
```diff
@@ -515,7 +515,7 @@ def execute(self, run_data, run_number, proposal, input_vars) -> 'Results':
                     log.warning(f"Skipping {name} because of missing dependencies: {', '.join(missing_deps)}")
                     # get error message from transient dependencies
                     dep_errors = [(dep, errors[dep]) for dep in missing_deps if dep in errors]
-                    if dep_errors:
+                    if dep_errors and not all(isinstance(e, Skip) for _, e in dep_errors):
                         errors[name] = DependencyError(dep_errors)
```
tmichela:
```suggestion
                    if dep_errors:
                        errors[name] = DependencyError(dep_errors)
                        if all(isinstance(e, Skip) for _, e in dep_errors):
                            errors[name] = Skip(str(errors[name]))
```

I suggest to keep building the dependency tree of skipped variables, to help understand why it was skipped.

PR #562, opened 2026-04-27
Fix(gui): delete column with tag filter active

tmichela:
fixes: If you delete a column while there are tag filters applied, all of the filtered out columns are saved as if you had manually hidden them, and the only way to see them again is to individually re-enable each one in the column settings dialog. This persists even after relaunching the GUI, because it's part of the settings.


takluyver:
Thanks, LGTM

PR #563, opened 2026-04-28
Allow selecting a run in the API with proposal & run numbers

takluyver:
Also fixes double-clicking a cell to preview data if you have a table with data from >1 proposal.


tmichela (commented):
Could you update the API documentation? Besides that this LGTM.

PR #566, opened 2026-04-30
GUI: Handle run_variables entries without run_info when loading the table

takluyver:
@turkot discovered it was unfortunately easy to get into a state where a proposal couldn't be opened in the GUI. In this case it was newly set up, so it could be deleted and started again, but this clearly isn't ideal. This should allow the GUI to cope with this condition.


tmichela:
LGTM

PR #567, opened 2026-05-05
damnit init: CLI to set up new DAMNIT folder

takluyver:
Various other commands which might currently implicitly set up a new database, like `damnit db-config`, will instead fail if not run in/on a DAMNIT folder. This also adds `damnit reprocess --in 11437 123`, since it's come up a few times that cd-ing to the location first is unintuitive.

I'd recommend reviewing commit-by-commit.

### DIFF
--- a/damnit/cli.py
+++ b/damnit/cli.py
@@ -2,6 +2,7 @@ import argparse
 
 import inspect
 import logging
+import re
 import sys
 import textwrap
 import traceback
@@ -42,6 +43,19 @@ def excepthook(exc_type, value, tb):
     import IPython
     IPython.start_ipython(argv=[], display_banner=False,
                           user_ns=target_frame.f_locals | target_frame.f_globals | {"__tb": lambda: print(tb_msg)})
+
+
+def get_existing_damnit_dir(proposal_or_dir: str | None):
+    if proposal_or_dir is None:
+        return None
+    elif (path := Path(proposal_or_dir)).is_dir():
+        return path
+    elif proposal_or_dir.isdigit():
+        proposal_name = f"p{int(proposal_or_dir):06d}"
+        return Path(find_proposal(proposal_name)) / "usr/Shared/amore"
+    else:
+        sys.exit(f"{proposal_or_dir} is not a proposal number or DAMNIT database directory")
+
 
 def handle_config_args(args, kv, converters={}):
     key = args.key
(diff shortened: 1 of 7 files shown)


damnit/cli.py
```diff
@@ -402,5 +526,111 @@ def main(argv=None):
...
+
+        def check_file(p: Path):
+            if not p.is_file():
+                sys.exit(f"{p} is not a file")
+            return p
+
+        context_file_src = user_vars_src = None
+        if like_dir := get_existing_damnit_dir(args.like):
+            context_file_src = check_file(like_dir / "context.py")
+            user_vars_src = check_file(like_dir / "damnit.py")
```
tmichela:
```suggestion
            user_vars_src = check_file(like_dir / "runs.sqlite")
```

damnit/cli.py
```diff
@@ -402,5 +526,111 @@ def main(argv=None):
...
+
+        if args.context is not None:
+            context_file_src = check_file(args.context)
```
tmichela, line 581:
```suggestion
            context_file_src = check_file(Path(args.context))
```

damnit/cli.py
```diff
@@ -402,5 +526,111 @@ def main(argv=None):
         elif args.migrate_subcmd == "intermediate-v1":
             migrate_intermediate_v1(db, Path.cwd(), args.dry_run)
 
+
+class InitSubcmd(Subcommand):
```
tmichela, line 530: it would be good to add tests for this new command in `test_cli.py`

damnit/cli.py
```diff
@@ -308,65 +185,250 @@ def main(argv=None):
...
+    help = "Run the DAMNIT combiner"
+
+    @staticmethod
+    def run(args: argparse.Namespace):
+        from .backend.combine import main
+        return main()
+
+
+class CombinerNowSubcmd(Subcommand):
+    name = "now",
```
tmichela:
```suggestion
    name = "now"
```

tmichela (changes requested):

damnit/cli.py
takluyver: Thanks, not sure what my brain was doing on that.

damnit/cli.py
takluyver: Thanks, I've done the wrapping with `type=Path` in the argument specification.

damnit/cli.py
takluyver: Good point, I've added a couple of tests now.

tmichela (approved):
I haven't checked in detail the whole refactoring, but I'll trust the test passing means no problem 🤞 

Random thought, I'm wondering if we can remove the migration.py script, it's unlikely we'll ever need it again.

LGTM.

PR #578, opened 2026-06-03
Minor GUI simplifications

takluyver:
Simplifying the GUI code a bit by enforcing a couple of restrictions

- The MainWindow is always opened with a DAMNIT directory (already the case in actual use, only the tests still opened an empty window)
- One window doesn't switch to a different DAMNIT directory - we open a new window instead.
- It always connects to Kafka, or the mock Kafka broker for the tests

There may be more simplifications possible. I also hope this will avoid bugs in rarely exercised code paths.

### DIFF
--- a/damnit/cli.py
+++ b/damnit/cli.py
@@ -132,10 +132,6 @@ class GuiSubcmd(Subcommand):
             help="Either a proposal number or a database directory."
         )
         parser.add_argument(
-            '--no-kafka', action='store_true',
-            help="Don't try connecting to XFEL's Kafka broker"
-        )
-        parser.add_argument(
             "--software-opengl", action="store_true",
             help="Force software OpenGL. Use this if displaying interactive Plotly plots shows a black screen."
                  "Active by default on Maxwell if not started on a display node."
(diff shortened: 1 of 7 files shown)


damnit/cli.py
```diff
@@ -131,10 +131,6 @@ def arguments(parser: argparse.ArgumentParser):
             'proposal_or_dir', nargs='?',
             help="Either a proposal number or a database directory."
         )
-        parser.add_argument(
-            '--no-kafka', action='store_true',
```
JamesWrigley, line 135: I always use this when running it locally 😛

takluyver: :melting_face: 

Would you be happy using setting the environment variable to a special value like `AMORE_BROKER=none` or `AMORE_BROKER=` instead of a command-line flag?

JamesWrigley: Not particularly but I can live with it. Why do you want to delete `--no-kafka`?

takluyver: I'm just trying to get rid of some of the less common states for the GUI, because they're room for bugs to hide. E.g. It would be easy for me to accidentally introduce a bug like using `update_agent` without checking that it's there, because I always do interactive tests on Maxwell with Kafka enabled.

Removing a command line flag was just a nice bonus. But as we do need to control it, and it needs to be passed through to `read_context`, an environment variable that's automatically inherited seems like a neater option.

JamesWrigley: I dunno if that's meaningfully simpler, but sounds ok :+1:

JamesWrigley (approved):
LGTM!

